PR #19, opened 2018-05-01
Set up CircleCI

benbromhead:
First cut of CircleCI for master. 
Includes build scripts that support building and releasing locally and from circleci


zegelin:
Not knowing much about Circle CI... looks good to me‽

PR #25, opened 2018-05-04
Enable DC deletion and Scale Up

liwang-pivotal:
1. Use class type to determine DC event type
2. Use Guava EventBus to receive and process DC event

### DIFF
--- a/operator/src/main/java/com/instaclustr/cassandra/operator/service/CassandraHealthCheckService.java
+++ b/operator/src/main/java/com/instaclustr/cassandra/operator/service/CassandraHealthCheckService.java
@@ -43,7 +43,6 @@ public class CassandraHealthCheckService extends AbstractScheduledService {
 
     private final CoreV1Api coreApi;
     private final Cache<DataCenterKey, DataCenter> dataCenterCache;
-    private final ControllerService controllerService;
 
     public interface StorageServiceMBean {
         List<String> getLiveNodes();
@@ -57,10 +56,9 @@ public class CassandraHealthCheckService extends AbstractScheduledService {
 
 
     @Inject
-    public CassandraHealthCheckService(final CoreV1Api coreApi,  final Cache<DataCenterKey, DataCenter> dataCenterCache, final ControllerService controllerService) {
+    public CassandraHealthCheckService(final CoreV1Api coreApi,  final Cache<DataCenterKey, DataCenter> dataCenterCache) {
         this.coreApi = coreApi;
         this.dataCenterCache = dataCenterCache;
-        this.controllerService = controllerService;
     }
 
 
@@ -71,12 +69,9 @@ public class CassandraHealthCheckService extends AbstractScheduledService {
         logger.debug("Checking health of cassandra instances...");
 
         for (final Map.Entry<DataCenterKey, DataCenter> cacheEntry : dataCenterCache.asMap().entrySet()) {
-        	
-        	logger.debug("Reconciling DataCenters...");
-        	
-        	controllerService.reconcileDataCenter(cacheEntry.getKey());
-        	
             final String labelSelector = String.format("cassandra-datacenter=%s", cacheEntry.getKey().name);
+
+            logger.debug("Checking health of cassandra instances...");
             
             final V1PodList podList = coreApi.listNamespacedPod(namespace, null, null, null, null, labelSelector, null, null, null, null);
 
--- a/operator/src/main/java/com/instaclustr/cassandra/operator/service/CassandraHealthCheckService.java
+++ b/operator/src/main/java/com/instaclustr/cassandra/operator/service/CassandraHealthCheckService.java
@@ -26,93 +26,93 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
 public class CassandraHealthCheckService extends AbstractScheduledService {
-	static final Logger logger = LoggerFactory.getLogger(CassandraHealthCheckService.class);
+    static final Logger logger = LoggerFactory.getLogger(CassandraHealthCheckService.class);
 
-	static final ObjectName STORAGE_SERVICE;
+    static final ObjectName STORAGE_SERVICE;
 
-	static {
-		try {
-			STORAGE_SERVICE = ObjectName.getInstance("org.apache.cassandra.db:type=StorageService");
+    static {
+        try {
+            STORAGE_SERVICE = ObjectName.getInstance("org.apache.cassandra.db:type=StorageService");
 
-		} catch (final MalformedObjectNameException e) {
-			throw new IllegalStateException(e);
-		}
-	}
+        } catch (final MalformedObjectNameException e) {
+            throw new IllegalStateException(e);
+        }
+    }
 
-	private final String namespace = "default";
+    private final String namespace = "default";
 
-	private final CoreV1Api coreApi;
-	private final Cache<DataCenterKey, DataCenter> dataCenterCache;
+    private final CoreV1Api coreApi;
+    private final Cache<DataCenterKey, DataCenter> dataCenterCache;
 
-	public interface StorageServiceMBean {
-		List<String> getLiveNodes();
-		List<String> getUnreachableNodes();
+    public interface StorageServiceMBean {
+        List<String> getLiveNodes();
+        List<String> getUnreachableNodes();
 
-		String getReleaseVersion();
-		String getSchemaVersion();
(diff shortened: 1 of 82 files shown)


liwang-pivotal:
@benbromhead CI tests failed with "ERROR: (gcloud.auth.activate-service-account) Could not read json file /root/gcp-key.json: No JSON object could be decoded"

benbromhead:
Sorry I've made a few changes to CI. Will investigate and resolve tomorrow :)

benbromhead:
Actually I just realized that CircleCI does not pass in ENV vars by default for PR builds to protect secrets from malicious PRs. 

I'm currently looking into a better workflow.

operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
```diff
@@ -72,8 +71,41 @@ void clusterEvent(final ClusterEvent event) {
     }
 
     @Subscribe
-    void dataCenterEvent(final DataCenterEvent event) {
-    	dataCenterEventQueue.add(event);
+    void dataCenterEvent(final DataCenterEvent event) throws Exception {
```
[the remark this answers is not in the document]
liwang-pivotal: @zegelin If using `DataCenterKey` as queue's element, when processing the queue how to determine the event type? DataCenterKey only has two fields - `name` & `namespace`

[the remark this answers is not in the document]
zegelin: Have a look at how the prometheus-operator works:
https://github.com/coreos/prometheus-operator/blob/master/pkg/prometheus/operator.go#L585

It follows the same pattern:
events -> resolve to a key -> put key on queue -> process key and reconcile state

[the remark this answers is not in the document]
liwang-pivotal: That's because `ResourceEventHandlerFuncs` can be used as a router, in our case we have to find a way to detect the event type

operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
```diff
@@ -121,25 +89,36 @@ void statefulSetEvent(final StatefulSetEvent event) {
...
+        	//processNextClusterEvent();
+        	//processNextSecretEvent();
+        	//processNextStatefulSetEvent();
+        }
+    }
+    
+    private void processNextDataCenterEvent() throws Exception {
+    	final DataCenterKey dataCenterKey = dataCenterQueue.take();
+        
+        if (dataCenterKey == POISON) return;
```
zegelin: This check needs to be in the main loop, rather than in a separate function. The intent of the `POISON` marker is to abort the loop.

operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
```diff
@@ -341,13 +326,18 @@ private static void deleteResource(final ApiCallable deleteCallable) throws ApiE
...
     	V1DeleteOptions deleteOptions = new V1DeleteOptions();
     	deleteOptions.setPropagationPolicy("Foreground");
     	
     	//Scale the statefulset down to zero (https://github.com/kubernetes/client-go/issues/91)
     	statefulSet.getSpec().setReplicas(0);
-    	createOrReplaceNamespaceStatefulSet(statefulSet);
+    	try {
+    		appsApi.replaceNamespacedStatefulSet(statefulSet.getMetadata().getName(), namespace, statefulSet, null);
+    	} catch (ApiException e) {
+    	    System.err.println("Exception when calling AppsV1beta2Api#replaceNamespacedStatefulSet");
```
zegelin: Use the logger

operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
```diff
@@ -341,13 +326,18 @@ private static void deleteResource(final ApiCallable deleteCallable) throws ApiE
...
+    	    e.printStackTrace();
```
zegelin: Rethrow?

operator/src/main/java/com/instaclustr/cassandra/operator/service/CassandraHealthCheckService.java
```diff
@@ -69,8 +71,13 @@ protected void runOneIteration() throws Exception {
         logger.debug("Checking health of cassandra instances...");
 
         for (final Map.Entry<DataCenterKey, DataCenter> cacheEntry : dataCenterCache.asMap().entrySet()) {
+        	
+        	logger.debug("Reconciling DataCenters...");
+        	
+        	controllerService.reconcileDataCenter(cacheEntry.getKey());
```
zegelin: Not sure if its the best idea to have the health check service, which just verifies the health of the Cassandra nodes, call out to the `ControllerService`.

liwang-pivotal: Created a seperate Repair service to repair missing components

operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
liwang-pivotal: add a comment here only for the format issue. Looks like it's related with my IDE. Reformatting the whole file will mess up with statefulset specs and other methods. might need some helps on this.

operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
```diff
@@ -220,14 +220,22 @@ private V1ConfigMap generateConfigMap(DataCenterKey dataCenterKey) throws IOExce
     private V1beta2StatefulSet generateStatefulSet(DataCenterKey dataCenterKey, V1ConfigMap configMap) {
     	final DataCenter dataCenter = dataCenterCache.getIfPresent(dataCenterKey);
     	
+    	int replicas = 0;
+    	try {
+    		replicas = dataCenter.getSpec().getReplicas().intValue();
+    	} catch (Exception e) {
+    		System.err.println("Invalid Replica Number.");
+    	    e.printStackTrace();
```
zegelin: Logging + rethrow

liwang-pivotal:
Reformatting the whole file will mess up with statefulset and other definitions. Looks like it's a issue related with my IDE. Might need some helps on this.

zegelin:
Which IDE are you using?

liwang-pivotal:
@zegelin Spring Tool Suite

zegelin:
Okay, I'm unfamiliar with that IDE.

I'll reformat the code later once marged.

liwang-pivotal:
@zegelin looks like I found a way to fix indentation. Should be good now.

benbromhead:
Otherwise lgtm 👍

liwang-pivotal:
@benbromhead @zegelin need your opinions on how to rename CassandraRepairService. My original thought is to include this feature into existing healthcheck service

benbromhead:
I would call it something like CassandraKubernetesResourceService or something?

zegelin:
Or `PeriodicReconciliationService`?

Though, I'm not quite sure why we need to periodically reconcile. If we don't recieve a modification event for a `DataCenter`, the local cache will be the same. All this periodic refresh service does is grabs every DC from the cache and tries to reconcile it, and if the cache hasn't been updated, the reconciliation wont achieve anything.

If the intention is to recreate missing objects (StatefulSets, Secrets, etc), then I had intended to have the operator watch for changes on those objects too. i.e. we get notified when the StatefulSet is deleted, rather than just polling every _n_ seconds to maybe recreate it.

liwang-pivotal:
@zegelin Yes, the intention is to prevent malicious deletion of primitive resources and recreate them automatically. As you said, to achieve that, there might be two ways: 1) periodically reconcile 2) create watcher for each resources. I think the advantage for 1) is it can reuse the reconcile code and thus don't require to maintain watchers for each resource, considering we might add more primitive resources in the future. However the down side of it might be the overhead for unnecessary reconcile and possible idle window to when repair is needed. I suggest we use current logic to have the feature enabled, and put optimization into our backlog.

liwang-pivotal:
@benbromhead CI test still seems to fail... Should I ignore it and merge, or I should sync with current master branch to fix this issue?

benbromhead:
if you rebase on master tests should pass

PR #29, opened 2018-05-11
Backup

benbromhead:
Implement standalone backup module
Additional CI changes for better testing
About time for a review


backup/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+        <rabbitmq.version>3.5.4</rabbitmq.version>
+        <thrift.version>0.9.3</thrift.version>
+        <azure.svc-mgmt.version>0.9.3</azure.svc-mgmt.version>
+        <azure.ad.auth.version>1.1.2</azure.ad.auth.version>
+        <azure.rsrc-mgmt.version>1.0.3-SNAPSHOT</azure.rsrc-mgmt.version>
+        <azure.storage.version>5.1.1</azure.storage.version>
+        <shiro.version>1.2.2</shiro.version>
+        <cassandra.driver.version>3.2.0</cassandra.driver.version>
+        <zendesk.client.version>0.5.4</zendesk.client.version>
+        <jira.version>0.5</jira.version>
```
zegelin: A bunch of these probably aren't needed any more.
Additionaly, i think deps like Guice and Guava should be pushed to the parent pom so that every jar has the same version (less stuff on disk).

backup/pom.xml
```diff
@@ -0,0 +1,136 @@
...
+            <groupId>org.mindrot</groupId>
+            <artifactId>jbcrypt</artifactId>
+            <version>0.3m</version>
+        </dependency>
+
+        <dependency>
+            <groupId>args4j</groupId>
+            <artifactId>args4j</artifactId>
+            <version>2.0.31</version>
+        </dependency>
```
zegelin: Okay for now, but I might work on switching to picocli (as the operator uses). It's a new lib with more features.

backup/src/main/java/com/instaclustr/backup/Application.java
```diff
@@ -0,0 +1,63 @@
+package com.instaclustr.backup;
```
zegelin: We probably should merge/refactor this with the operator code at some point, since it uses the same services + servicemanager + guice model

backup/src/main/java/com/instaclustr/backup/CommonBackupArguments.java
```diff
@@ -0,0 +1,115 @@
+package com.instaclustr.backup;
```
zegelin: Whats the dif between `CommonBackupArguments`, `BaseBackupArguments` and `BackupArguments`?

backup/src/main/java/com/instaclustr/backup/downloader/AzureDownloader.java
```diff
@@ -0,0 +1,77 @@
...
+    @Override
+    public RemoteObjectReference objectKeyToRemoteReference(final Path objectKey) throws StorageException, URISyntaxException {
+        final String path = resolveRemotePath(objectKey);
+        return new AzureRemoteObjectReference(objectKey, path, blobContainer.getBlockBlobReference(path));
+    }
+
+    @Override
+    public void downloadFile(final Path localPath, final RemoteObjectReference object) throws Exception {
+        final CloudBlockBlob blob = ((AzureRemoteObjectReference) object).blob;
+        final File localFilePath = localPath.toFile();
```
zegelin: Mixing `Path`s and `File`s.

backup/src/main/java/com/instaclustr/backup/task/BackupTask.java
```diff
@@ -0,0 +1,339 @@
...
+                    matcher.matches();
+                    return matcher.group(SSTABLE_GENERATION_IDX);
+                }));
+    }
+
+    @VisibleForTesting
+    public static String calculateChecksum(final Path filePath) throws IOException {
+        try(final FileInputStream fileInputStream = new FileInputStream(filePath.toFile())) {
+
+            final FileChannel fileChannel = fileInputStream.getChannel();
```
zegelin: Mixing `FileChannel`s and `InputStream`s

backup/src/main/java/com/instaclustr/backup/task/TaskFactory.java
```diff
@@ -0,0 +1,5 @@
+package com.instaclustr.backup.task;
+
+public interface TaskFactory {
+    BackupTask backupTask();
```
zegelin: Is this factory used? Perhaps it should also create `RestoreTask`s?

backup/src/main/java/com/instaclustr/backup/uploader/LocalFileSnapShotUploader.java
```diff
@@ -0,0 +1,55 @@
...
+    }
+
+    @Override
+    public RemoteObjectReference objectKeyToRemoteReference(final Path objectKey) throws Exception {
+        return new LocalFileObjectReference(objectKey, resolveRemotePath(objectKey));
+    }
+
+    @Override
+    public FreshenResult freshenRemoteObject(final RemoteObjectReference object) throws Exception {
+        final File fullRemoteObject = resolveFullRemoteObjectPath(object).toFile();
```
zegelin: Mixing `File`s and `Path`s again.

backup/src/main/java/com/instaclustr/backup/task/TaskFactory.java
benbromhead: Deleted

benbromhead:
I'm actually going to keep as much of the deps separate from the operator, as we have some customers who are going to using the backup application as a standalone backup capability for Cassandra. 

Most likely the backup module will be pulled into its own separate repo as a standalone project. When that happens we'll need to make sure it's  still easily "embeddable".

backup/src/main/java/com/instaclustr/backup/CommonBackupArguments.java
benbromhead: This is a relic from our MS backup component that also supports Commitlog archiving so it has a different set of args for the commitlog entry point. I've left it this way as "at some point" we'll re-introduce commitlog archiving, but it can easily be cleaned up

backup/src/main/java/com/instaclustr/backup/Application.java
benbromhead: See comment on likelihood that backup will be moved into a separate repo

backup/pom.xml
benbromhead: 👍 Will update when that change lands

backup/pom.xml
benbromhead: 👍

benbromhead:
@lyubentt feel free to join in on this one

benbromhead:
WRT to guice, it kind of was doubling up with the args (args object gets passed around instead of an injector) and for such a small set of dependencies it didn't make a ton of sense. 

Given this is going to be a fairly standalone component, I think we'll just leave as is.

Backup scheduling, yup I'll take that out as it was not my intention to have the module doing the scheduling itself. If we need it, we can add it back pretty easily.

backup/src/main/java/com/instaclustr/backup/task/BackupTask.java
benbromhead: Rereading the whole thing, it needs a bit of a rework. Lemme get back to you on that one

backup/src/main/java/com/instaclustr/backup/task/BackupTask.java
```diff
@@ -229,31 +233,30 @@ public void run() {
...
     @VisibleForTesting
     public static String calculateChecksum(final Path filePath) throws IOException {
-        try(final FileInputStream fileInputStream = new FileInputStream(filePath.toFile())) {
-
-            final FileChannel fileChannel = fileInputStream.getChannel();
+        try(final FileChannel fileChannel = FileChannel.open(filePath)) {
 
-            long bytesStart;
+            int bytesStart;
             int bytesPerChecksum = 10 * 1024 * 1024;
```
zegelin: Maybe make this a constant?

backup/src/main/java/com/instaclustr/backup/task/BackupTask.java
```diff
@@ -229,31 +233,30 @@ public void run() {
...
 
             // Get last 10 megabytes of file to use for checksum
             if (fileChannel.size() >= bytesPerChecksum) {
-                bytesStart = fileChannel.size() - bytesPerChecksum;
+                bytesStart = toIntExact(fileChannel.size()) - bytesPerChecksum;
```
zegelin: Is this to avoid overflow?

zegelin:
It skipped my mind that we'd like to keep the backup stuff separate so that it can be run standalone.

Looks good to me.

PR #48, opened 2018-06-01
Temporary fix for Unprocessable Entity issue

liwang-pivotal:
Looks like only `service` replace call will cause `Unprocessable Entity` exception and make watch service fail. Instead, `configmap` and `statefulset` resource replace call won't break watch threads.
So the solution is temporarily disable `replaceNamespacedServiceCall`, since currently services are not required to be updated.


liwang-pivotal:
Rebased & cleaned commits. Now is PR #51

PR #50, opened 2018-06-04
Minor tweaks

zegelin:
Minor refactoring and bug fixes pending Sidecar module.

- Splits some utility classes into Common module.
- Fixes issue where CassandraHealthCheckService would fire events even if the status hadn't changed.
- Fixes issue where each K8s "client" would create a whole new ApiClient, etc -- they are now `@Singletons`
	- This also fixes the WatchService refreshing every 10 seconds rather than the timeout (currently 1min)


zegelin:
This is dependent on PR #49, hence the duplicate commits. Please review commit 
`6d1d071`

java/operator/src/main/resources/com/instaclustr/test-dc.yaml
```diff
@@ -4,13 +4,17 @@ metadata:
   name: test-dc
 spec:
   replicas: 6
-  image: instaclustr/k8s-cassandra
+  version: 3.11.2
+  image: instaclustr/k8s-cassandra:3.11.2
```
benbromhead: can we change this to the image name/tag used by default with the build scripts and stored in gcr.io ?

zegelin: Yep, will change that. I'll also remove `version` -- that was me experimenting with things, but I eventually realised that its just far easier to allow a full image string and just have a big fat warning that the operator only works with C* images built with the operator in mind.

PR #53, opened 2018-06-07
Remove PVC & PV when scaling down or DC get deleted

liwang-pivotal:
Now when a pod get deleted (scaling down or DC delete event), it's Persistent Volume and Persistent Volume Claim will also got deleted.


liwang-pivotal:
Also added back `scale down to 0` code when deleting StatefulSet.

PR #56, opened 2018-06-07
Disable version check if running on minikube

liwang-pivotal:
minikube doesn't provide `major` and `minor` info by default. Originally to run operator on minikube, one should modify cassadran-operator.sh to add `--no-version-check`, otherwise it will throw `NumberFormatException`. Now `K8sVersionValidator` should be able to skip version check on minikube.


zegelin:
What version of Minikube are you running? I'm running v0.26.0 and don't see this issue.

Additionally, you shouldn't need to modify `cassandra-operator.sh` to pass command line arguments. It's a wrapper script, and any arguments passed to the script will be passed to the operator, i.e.:
```/opt/bin/cassandra-operator --no-version-check'''

zegelin:
Actually, I just realised that a fix for `cassandra-operator.sh` is present in #50.
There was a missing escape that prevents any arguments being passed into the JVM.

liwang-pivotal:
Ha, that's a better way!

PR #60, opened 2018-06-19
Enable continuation of scale-down util target replicas is hit

liwang-pivotal:
Previously, the operator was limited to scale cluster down by one pod at a time unless another `DataCenterEvent` is triggered. Now the operator is able to fire another `DataCenterEvent` once it confirms the previous decommission is finished.

### DIFF
--- a/doc/op_guide.md
+++ b/doc/op_guide.md
@@ -11,7 +11,7 @@ minikube start --cpus 4 --memory 4096 --kubernetes-version v1.9.4
 eval $(minikube docker-env)
 
 #Build the operator
-mvn -f java clean package
+mvn -f java/pom.xml clean package
 
 #Build the required docker images
 ./buildenv/build-all
--- a/examples/common/rbac-bundle.yaml
+++ b/examples/common/rbac-bundle.yaml
@@ -31,10 +31,12 @@ rules:
 - apiGroups:
   - stable.instaclustr.com
   resources:
-  - cluster
-  - clusters
-  - datacenter
-  - datacenters
+  - cassandra-cluster
+  - cassandra-clusters
+  - cassandra-datacenter
+  - cassandra-datacenters
+  - cassandra-backup
+  - cassandra-backups
   verbs:
   - '*'
 - apiGroups:
@@ -101,7 +103,4 @@ spec:
           requests:
             cpu: 100m
             memory: 50Mi
-      securityContext:
-        runAsNonRoot: true
-        runAsUser: 65534
       serviceAccountName: cassandra-operator
--- a/examples/common/test.yaml
+++ b/examples/common/test.yaml
@@ -6,3 +6,14 @@ spec:
   replicas: 3
   image: "gcr.io/cassandra-operator/cassandra:latest"
   imagePullPolicy: IfNotPresent
+  resources:
+    limits:
+      memory: 512Mi
+    requests:
+      memory: 512Mi
+  dataVolumeClaim:
+    accessModes:
+      - ReadWriteOnce
+    resources:
+      requests:
+        storage: 100Mi


liwang-pivotal:
@benbromhead If controller is restarted accidentally before [firing](https://github.com/liwang-pivotal/cassandra-operator/blob/2a7c9bf0cc20331e8344b572af7453da38eb56f7/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java#L393) the `DataCenterEvent`, since the `scale` command has already been [issued](https://github.com/liwang-pivotal/cassandra-operator/blob/2a7c9bf0cc20331e8344b572af7453da38eb56f7/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java#L370), the pod removal process will continue. Every start of a new controller will receive a initial `DataCenterEvent` to trigger the reconcile, so basically it's the same as no restart of the controller.

java/operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java
```diff
@@ -374,6 +374,24 @@ private void replaceStatefulSet(final DataCenter dataCenter, final V1beta2Statef
             } catch (final Exception e) {
                 logger.warn("Failed to delete PersistentVolume & PersistentVolumeClaim. Reason is: ", e);
             }
+            
+            // waiting for the termination of above pod
+            while (true) {
```
benbromhead: Do we want to spin loop on this for ever? Or should we just give up at some point?

Otherwise LGTM

liwang-pivotal: Ah, I got what you worried about. If there is a restart of the controller, this [condition](https://github.com/liwang-pivotal/cassandra-operator/blob/2a7c9bf0cc20331e8344b572af7453da38eb56f7/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java#L383) will be satisfied and then exit the loop. I know the loop here looks ugly but can't find a better way to wait for the finish of scale-down above.

zegelin:
Instead of spin-looping, lets react to the StatefulSets being updated/modified once its status changes.
There's already a watch over StatefulSets, but the controller currently doesn't act on the events.

ie.
controller updates statefulset -> statefulset then scales down -> statefulset status is updated -> triggers watch event -> controller runs reconcile -> continues with next scale down.

Still need to think of a way to "de-bounce" events, since the act of modifying the `replicas` field of the statefulset would trigger an event -- ie. our own modifications trigger events, which we should ignore. Okay for now, but probably will eventually get expensive.

Also currently the StatefulSet watch is for every SS in the namespace, should have a label-matcher -- new my TODO in `StatefulSetWatchModule`.

liwang-pivotal:
@zegelin I agree with the workflow you mentioned except that - the statefulSet watch will receive the event once the scale down is triggered. Without a loop it can't tell if the scale down is finished.

zegelin:
It will also receive an event once the scale down is finished.

liwang-pivotal:
@zegelin Okay I see. Just did a test, there are modified event fired once finished. How can we filter those modified statefulset events specific to `scale down & finish`? Looks like currently we only can tell it's a modified event.

zegelin:
It shouldn't matter. Reconcile should be idempotent -- ie. it can be called multiple times and will only change things when possible/necessary.

If reconcile is triggered when the SS hasn't finished scaling down, it should just not do anything.

liwang-pivotal:
@zegelin The reconcile will cause JMX connection scan which might be expensive, especially every one pod scale down will cause 4 modified events during the test. 
Also another issue we talked about is popped up on [code](https://github.com/liwang-pivotal/cassandra-operator/blob/2a7c9bf0cc20331e8344b572af7453da38eb56f7/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/ControllerService.java#L355-L361) here - `CassandraConnection` will be unable to detect operation mode and throw out errors.

zegelin:
@liwang-pivotal right, well lets fix the JMX slowness, etc separately.

zegelin:
I might also look into a way of somehow "coalescing" multiple queued changes into one.

PR #67, opened 2018-07-09
Adding imagePullPolicy: Always as the default

temujin9:
The issues I've been experiencing with helm appear to have been an old cached Docker image. Adding the following changes made the helm packages install the latest version correctly, moving past the issue.


zegelin:
@benbromhead is this still something that needs to merge?

benbromhead:
On reflection, I don't think it makes sense to always pull the image by default. Once the project moves to beta, we will adopt versioning as part of the docker image release process so users have explicit control over what they are running.

PR #68, opened 2018-07-11
Adds first-cut of the ConcatenatedYamlConfigurationLoader

zegelin:


benbromhead:
LGTM

Note that we will need to document the fact that duplicate key relies on implied behavior from snakeyaml (e.g. it doesn't complain and last key wins). 

This is correct according to yaml spec (keys should be unique, but duplicates are handled by the application not the parser)

PR #77, opened 2018-08-02
Fix some Helm default values

benbromhead:
Use a better pull policy
Change helm to use images built off master


zegelin:
Looks like this duplicates some of #67

PR #90, opened 2018-08-22
permissions to delete pvc

benbromhead:
Proposed fix for  #89

### DIFF
--- a/helm/cassandra-operator/templates/clusterrole.yaml
+++ b/helm/cassandra-operator/templates/clusterrole.yaml
@@ -50,7 +50,8 @@ rules:
   resources:
   - services
   - endpoints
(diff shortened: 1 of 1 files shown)


liwang-pivotal:
@benbromhead I've tested this patch on PKS, but got the same exception. I've cleaned the env and tried with the latest `ClusterRole`. Also I changed back to `cluster-admin` role, it works well. So now I'm pretty sure it's a `ClusterRole` issue but not only for `persistentvolumeclaims`.

liwang-pivotal:
@benbromhead Poked around and found `persistentvolumes` should be added as well : ) Then I confirmed we're able to delete cluster.

benbromhead:
@liwang-pivotal changes added

PR #91, opened 2018-08-24
Refactor of controller and configuration management

zegelin:
Major things of note:

- Renamed `ControllerService` to `OperatorService`, since the project is an _operator_ not a _controller_ ;)
- Refactored `OperatorService` into a service that just manages queueing watched changes -- eventually we could have a small delay and coalesce events for the same DC key, avoiding unnecissary work.
- `DataCenterReconciliationController` and `DataCenterDeletionController` now are responsible for reconcilliation and deletion, respectivly. This cleans up a lot of code for the creation/update of all the resources since a controller is specific to a _single DC_ so there is no-longer `dataCenter` parameters everywhere (yay).
- Config wise, everything now has a `.d` directory too: `cassandra.yaml.d`, `cassandra.env.d`, `jvm.options.d`, etc.
- Config map volumes are now mounted into `/tmp/<some name>` and their contents copied and overlayed onto `/etc/cassandra`. This allows for multiple config maps (current operator defined + user overrides) to modify anything under `/etc/cassandra`. Yes this breaks at-runtime config map updates, but C* doesn't reload its config without a restart anyway...
- Rather than having the default config files (not sure still what should be "defaults") baked into the operator and always copied verbatim into a config map, they're now baked into the docker image.
- The operator handles operator defined config and user overloads in the same way.

And a number of other small things I've probably forgotten.

I should mention that this isn't yet ready for a merge. I wanted to get eyes on it.


docker/cassandra/Dockerfile
```diff
@@ -6,13 +6,19 @@ COPY install-cassandra /tmp/install-cassandra
 
 RUN /tmp/install-cassandra ${cassandra_version}
 
+ADD https://github.com/zegelin/cassandra-exporter/releases/download/v0.9.0/cassandra-exporter-agent-0.9.0.jar /usr/share/cassandra/agents
```
zegelin: Oops, this slipped in.

docker/cassandra/cassandra.sh
```diff
@@ -1,45 +1,44 @@
...
 
 # provides hints to the JIT compiler
-JVM_OPTS="${JVM_OPTS} -XX:CompileCommandFile=/usr/share/cassandra/hotspot_compiler"
+JVM_OPTS="${JVM_OPTS} -XX:CompileCommandFile=${CASSANDRA_HOME}/hotspot_compiler"
 
-# add the jamm javaagent
+# add the jamm agent
 JVM_OPTS="${JVM_OPTS} -javaagent:${CASSANDRA_HOME}/agents/jamm-0.3.0.jar"
 
+## Prometheus exporter agent
```
zegelin: Commented out, as this will instead be added as an operator config override.

docker/cassandra/cassandra.sh
```diff
@@ -1,45 +1,44 @@
...
-. ${CASSANDRA_CONF}/cassandra-env.sh
+# source additional environment settings
+for env_file in "${CASSANDRA_CONF}/cassandra-env.sh" "${CASSANDRA_CONF}/cassandra-env.sh.d"/*
+do
+    . "${env_file}"
+done
 
 
-exec -a cassandra /usr/bin/java \
+exec /usr/bin/java \
```
zegelin: Need to revert this.

docker/cassandra/default-config/cassandra.yaml
```diff
@@ -1,13 +1,6 @@
-# Cassandra storage config YAML
+data_file_directories:
```
zegelin: Not sure what we should consider as the "defaults" here.

Setting the directories -- yes, but other things? I guess whatever the out-of-box settings are...

zegelin: The way C* handles defaults is complex: with the some config options having defaults in the `Config` class, others requiring something to be set in the YAML for C* to start, and others where the code default and the YAML default differ...

I might leave the package distro `cassandra.yaml` in `/etc/cassandra` and everything else can be done as overides.

docker/cassandra/default-config/jvm.options
```diff
@@ -0,0 +1,42 @@
+-Dcassandra.jmx.remote.port=7199
```
zegelin: Seemed like a bunch of good defaults.

Specifically excluded is the GC settings -- the operator should set this based on the container memory allocation.

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -0,0 +1,414 @@
...
+                );
+
+        // if the StatefulSet doesn't exist, create it. Otherwise scale it safely
+        K8sResourceUtils.createOrReplaceResource(
+                () -> appsApi.createNamespacedStatefulSet(statefulSet.getMetadata().getNamespace(), statefulSet, null),
+                () -> replaceStatefulSet(statefulSet)
+        );
+    }
+
+//    private V1ConfigMap createOrReplaceCassandraConfigMap(final DataCenter dataCenter) throws IOException, ApiException {
```
zegelin: This is still WIP.

docker/cassandra/default-config/logback.xml
```diff
@@ -0,0 +1,78 @@
...
+    </appender>
+
+    <!-- Uncomment bellow and corresponding appender-ref to activate logback metrics
+    <appender name="LogbackMetrics" class="com.codahale.metrics.logback.InstrumentedAppender" />
+     -->
+
+    <root level="INFO">
+        <appender-ref ref="SYSTEMLOG" />
+        <appender-ref ref="STDOUT" />
+        <appender-ref ref="ASYNCDEBUGLOG" /> <!-- Comment this line to disable debug.log -->
```
benbromhead: Can this be overridden by the user?

zegelin: Not yet. I'm still thinking about how best to handle logs.

zegelin: Actually, scratch that. Yes it is, but you have to overwrite the whole file.

A configmap with a logback xml, and then a configmap volume mount with

```
key: "logback-xml"
path: "logback.xml"
```

should work

docker/cassandra/default-config/jvm.options
benbromhead: The operator should probably have exclusive control of setting jmx ports / configs, as the operator will always need to be able to do things to C* via JMX (e.g. decommission) unless of course, we shift that to the sidecar.

zegelin: Agree. I moved all the default configs back into the image because it seemed cleaner as the operator doesn't have to copy huge files and make a mess in a configmap for every deployment. Additionally it means that the image runs fine standalone, which helped me test a few things.

Having this baked-in with a known port number has no issues AFAIK -- but we could always drop in a jvm.options.d file with an explicit port.

Either way, not sure how we can totally prevent people from changing this port number since we allow custom jvm.options/env override. The result would be that the operator doesn't work and the fix is "don't do that".

docker/cassandra/cassandra.sh
```diff
@@ -32,15 +22,21 @@ JVM_OPTS="${JVM_OPTS} -Djava.library.path=${CASSANDRA_HOME}/lib/sigar-bin"
...
 
+# source additional environment settings
+for env_file in "${CASSANDRA_CONF}/cassandra-env.sh" "${CASSANDRA_CONF}/cassandra-env.sh.d"/*.sh
+do
+    . "${env_file}"
+done
 
 exec -a cassandra /usr/bin/java \
     -cp "${CASSANDRA_CLASSPATH}" \
+    -ea \   # beacuse turing this off is a bad, bad idea
```
rs017991: That comment may be defeating the escape of the newline

zegelin: @rs017991 yep. thanks for pointing that out.
I've fixed this in #92, but I'll back-port it.

zegelin:
@benbromhead Updated documentation

PR #101, opened 2018-10-08
Set kernel caps properly

benbromhead:
Support setting required kernel capabilites and limits for Cassandra to work good.


benbromhead:
Have to set the max_map_count via a privileged init container. For the moment we just one-shot the sidecar container with a different command. This is to get around the read-only proc fs in non-privileged containers.

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -255,6 +261,15 @@ private void createOrReplaceStateNodesStatefulSet(final Iterable<ConfigMapVolume
         );
     }
 
+    private V1Container fileLimiteInit() {
```
zegelin: Olde Syse Limite?

benbromhead: Very Olde and Limite

PR #103, opened 2018-10-10
Continue scaling when replica count changes by > 1

benbromhead:
The operator now supports replica count changes greater than increments of 1
The operator will also react to decommissioned nodes

### DIFF
--- a/java/backup/pom.xml
+++ b/java/backup/pom.xml
@@ -32,7 +32,6 @@
             <version>24.1-jre</version>
         </dependency>
 
-        <!--TODO: change to BOM style to only include s3 deps -->
         <dependency>
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk</artifactId>
--- a/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/OperatorService.java
+++ b/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/OperatorService.java
@@ -75,7 +75,7 @@ public class OperatorService extends AbstractExecutionThreadService {
 
     @Subscribe
     void handleCassandraNodeOperationModeChangedEvent(final CassandraNodeStatusChangedEvent event) {
-        logger.trace("Received CassandraNodeStatusChangedEvent {}.", event);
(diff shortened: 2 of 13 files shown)


benbromhead:
Also includes support for watching custom namespaces (just one) per operator instance.

java/backup/pom.xml
```diff
@@ -32,7 +32,6 @@
             <version>24.1-jre</version>
         </dependency>
 
-        <!--TODO: change to BOM style to only include s3 deps -->
```
zegelin: This TODO still applies though, right? We don't really want to pull in all the other frameworks...

java/operator/src/main/java/com/instaclustr/cassandra/operator/service/OperatorService.java
```diff
@@ -48,13 +48,19 @@ void handleDataCenterEvent(final DataCenterWatchEvent event) {
...
+        // TODO: handle updated/deleted secrets - currently we don't care about this
     }
 
     @Subscribe
     void handleStatefulSetEvent(final StatefulSetWatchEvent event) {
         logger.trace("Received StatefulSetWatchEvent {}.", event);
-        // TODO
+
+        // Trigger a dc reconciliation event if changes to the stateful set has finished.
+        if(event.statefulSet.getStatus().getReplicas().equals(event.statefulSet.getStatus().getReadyReplicas()) && event.statefulSet.getStatus().getCurrentReplicas().equals(event.statefulSet.getStatus().getReplicas())) {
```
zegelin: I wonder what code style enforcement tools exist to automatically run on commit.

It'd be nice to enforce the following, for consistency:
- `final` everywhere
- whitespace after `if` and before parentheses
- always use braces, even for one-liner if statements

java/backup/pom.xml
benbromhead: Ah, yup it does. I'll put it back in.  

I did give it a go for 5 minutes, but stuff broke and I didn't want to fight maven and dependencies. So will look at soon.

java/operator/src/main/java/com/instaclustr/cassandra/operator/service/OperatorService.java
benbromhead: Fixed some if statements. Created #106 to address

PR #104, opened 2018-10-12
Fix cloud backups

benbromhead:
Fixed cloud backups to support new user-defined config map volumes.
Fixed Helm packages to support backups/user defined config map volumes.
Better error handling for backup API calls.

### DIFF
--- a/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/BackupControllerService.java
+++ b/java/operator/src/main/java/com/instaclustr/cassandra/operator/service/BackupControllerService.java
@@ -131,22 +131,22 @@ public class BackupControllerService extends AbstractExecutionThreadService {
                 .map(x -> x.getKey() + "=" + x.getValue())
                 .collect(Collectors.joining(","));
 
+        BackupSpec backupSpec = backup.getSpec();
+
         if(coreApi.listNamespacedPod(backup.getMetadata().getNamespace(), null, null,
                 null, null, labels, null, null, null, null)
                 .getItems()
                 .parallelStream()
                 .map(x -> callBackupApi(x, backup))
                 .anyMatch(e -> !e)) {
-            BackupSpec backupSpec = backup.getSpec();
             backupSpec.setStatus("ERROR");
-            backup.setSpec(backupSpec);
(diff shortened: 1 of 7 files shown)


java/operator/src/main/java/com/instaclustr/cassandra/operator/service/BackupControllerService.java
```diff
@@ -128,9 +137,16 @@ private void createOrReplaceBackup(final Backup backup) throws ApiException, Unk
                 .parallelStream()
                 .map(x -> callBackupApi(x, backup))
                 .anyMatch(e -> !e)) {
-            //TODO: Set backup status to PROCESSED
+            BackupSpec backupSpec = backup.getSpec();
+            backupSpec.setStatus("ERROR");
+            backup.setSpec(backupSpec);
+            customObjectsApi.patchNamespacedCustomObject("monitoring.coreos.com", "v1", backup.getMetadata().getNamespace(), "cassandra-backups", backup.getMetadata().getName(), backup);
```
zegelin: `monitoring.coreos.com`?

java/operator/src/main/java/com/instaclustr/cassandra/operator/service/BackupControllerService.java
```diff
@@ -128,9 +137,16 @@ private void createOrReplaceBackup(final Backup backup) throws ApiException, Unk
...
+        } else {
+            BackupSpec backupSpec = backup.getSpec();
```
zegelin: I think these two if branches can be consolidated. The only difference seems to be the status value -- "ERROR" vs "PROCESSED"?

java/operator/src/main/java/com/instaclustr/cassandra/operator/service/BackupControllerService.java
benbromhead: that would explain the missing status updates

java/operator/src/main/java/com/instaclustr/cassandra/operator/service/BackupControllerService.java
benbromhead: Consolidated

PR #108, opened 2018-10-16
support secret volumes

benbromhead:


java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -200,6 +203,18 @@ private void createOrReplaceStateNodesStatefulSet(final Iterable<ConfigMapVolume
             );
         }
 
+        if (secretVolumeSource != null) {
+            cassandraContainer.addVolumeMountsItem(new V1VolumeMount()
+                    .name("user-secret-volume")
+                    .mountPath("/tmp/user-config"));
```
zegelin: Not sure what happens if the user config mount is configured as well. AFAIK, you can only mount one thing at a mount point? Maybe?

Instead, maybe look into refactoring `ConfigMapVolumeMount` into `ConfigMapVolumeMount` and `SecretVolumeMount` with some kinda interface/base class that provides mount-point. Then get `createOrReplaceStateNodesStatefulSet(...)` to take an interable of that interface/base and extend the for loop above to work on both types.

That way, the mountpoint path will be passed into the entry-point script of the image, just like the config map volumes.

benbromhead: I've just changed it to a different mount point. Secrets generally don't need to get put in the same config folder as we don't need to shuffle them around, so for the moment let's just use a different directory altogether.

benbromhead:
@zegelin ping

PR #111, opened 2018-10-22
Fix vulnerability associated with commons compress lib

benbromhead:
CVE-2018-11771 More information
moderate severity
Vulnerable versions: < 1.18
Patched version: 1.18


benbromhead:
Note vulnerability requires an attacker manipulating your backed up C* files... in which you would have bigger problems. 

A corrupted backup may also trigger this

Issue #116, opened 2018-11-07
Allow running on unprivileged clusters


zegelin:
Trying to run the operator/cassandra on an unprivileged cluster results in an error.

```create Pod cassandra-test-cluster-cassandra-0 in StatefulSet cassandra-test-cluster-cassandra failed error: Pod "cassandra-test-cluster-cassandra-0" is invalid: spec.initContainers[0].securityContext.privileged: Forbidden: disallowed by cluster policy ```

Maybe we can make the init container optional?

benbromhead:
We need to run a privileged container to configure certain OS options that will allow Cassandra to run in a reliable manner. 

We can certainly make it optional, but it will make Cassandra more likely to break.

alourie:
@zegelin following from #208, will work on this soon. Will restore the `privelegedSupported` functionality and whatever else will be needed here.

alourie:
`privilegedSuppored` functionality is restored by #218

PR #118, opened 2018-11-13
Optional privileged

benbromhead:
Address #116 and allow users to run on clusters that don't support privileged containers.

### DIFF
--- a/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
+++ b/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
@@ -284,11 +284,11 @@ public class DataCenterReconciliationController {
 
     private V1Container fileLimitInit() {
         return new V1Container()
-                .securityContext(new V1SecurityContext().privileged(true))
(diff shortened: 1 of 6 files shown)


java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -284,11 +284,11 @@ private void createOrReplaceStateNodesStatefulSet(final Iterable<ConfigMapVolume
 
     private V1Container fileLimitInit() {
         return new V1Container()
-                .securityContext(new V1SecurityContext().privileged(true))
+                .securityContext(new V1SecurityContext().privileged(dataCenterSpec.getPrivilegedSupported()))
```
zegelin: At some point we should do better validation of the CRD object, since this could be null.

benbromhead: I'm going to mark this as a required field in the CRD. It's not super defensive at the operator level, but at least it should catch the vast majority of NPE this could generate.

PR #127, opened 2019-01-11
Feature/pullsecret support

### LINKED ISSUE #126: Support using imagePullSecrets
I'm testing this on IBM Cloud's Kubernetes service and would like to use `imagePullSecrets` for added security and ease of testing -- I can easily push dev builds to IBM Clould's container registry and test outside minikube. I did some preliminary work on making this happen but wanted to submit a PR before going too far.

cin:
This not only makes it so the operator sets up the imagePullSecret properly, but it also allows the build/release scripts to work outside of `gcloud`. I'm simply exporting `REPO=myreponame/mynamespace` and `DONT_LOGIN=true` before running build/release commands.

Resolves #126

### DIFF
--- a/helm/cassandra/values.yaml
+++ b/helm/cassandra/values.yaml
@@ -5,7 +5,7 @@
 replicaCount: 3
 
 image:
-  cassandraRepository: gcr.io/cassandra-operator/cassandra
+  cassandraRepository: gcr.io/cassandra-operator/cassandra-dev
   sidecarRepository: gcr.io/cassandra-operator/cassandra-sidecar-dev
   cassandraTag: 3.11.3
   sidecarTag: latest
--- a/helm/cassandra/values.yaml
+++ b/helm/cassandra/values.yaml
@@ -16,9 +16,9 @@ privilegedSupported: true
 
 resources:
   limits:
-    memory: 1024Mi
+    memory: 512Mi
   requests:
-    memory: 1024Mi
+    memory: 512Mi
 dataVolumeClaim:
   accessModes:
     - ReadWriteOnce
--- a/helm/cassandra/values.yaml
+++ b/helm/cassandra/values.yaml
@@ -11,15 +11,14 @@ image:
   sidecarTag: latest
 
 imagePullPolicy: IfNotPresent
-imagePullSecret: ""
 
 privilegedSupported: true
 
 resources:
   limits:
-    memory: 512Mi
+    memory: 1024Mi
   requests:
-    memory: 512Mi
+    memory: 1024Mi
 dataVolumeClaim:
   accessModes:
     - ReadWriteOnce
--- a/helm/cassandra/values.yaml
+++ b/helm/cassandra/values.yaml
@@ -11,6 +11,7 @@ image:
   sidecarTag: latest
 
 imagePullPolicy: IfNotPresent
+imagePullSecret: ""
 
 privilegedSupported: true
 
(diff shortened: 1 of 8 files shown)


benbromhead:
Thanks @cin I will review next week! cc @zegelin

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -116,6 +116,17 @@ private ConfigMapVolumeMount(final String name, final String mountPath, final V1
         }
     }
 
+    private V1PodSpec createPodSpec() {
+        V1PodSpec podSpec = new V1PodSpec();
```
zegelin: Please follow the code style of this project and make all local variables `final` unless they need to be reassigned.

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -116,6 +116,17 @@ private ConfigMapVolumeMount(final String name, final String mountPath, final V1
         }
     }
 
+    private V1PodSpec createPodSpec() {
```
zegelin: This looks like a partial refactor. This PR puts creating the `V1PodSpec` instance and assigning optional secrets, but leaves the rest of the pod spec creation code in `createOrReplaceStateNodesStatefulSet`.

It should be all in the same spot. Either we pull out the entire pod spec creation code into this function, or we remove this function and keep the code in `createOrReplaceStateNodesStatefulSet`.

zegelin: Note that we're already doing optional changes to the pod spec in `createOrReplaceStateNodesStatefulSet`.

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
```diff
@@ -116,6 +116,17 @@ private ConfigMapVolumeMount(final String name, final String mountPath, final V1
...
+        V1PodSpec podSpec = new V1PodSpec();
+        String secret = dataCenterSpec.getImagePullSecret();
+        if (secret != null && secret.length() > 0) {
```
zegelin: This project has Google Guava as a dependency which provides `Strings.isNullOrEmpty(...)`.

helm/cassandra/values.yaml
```diff
@@ -5,12 +5,13 @@
 replicaCount: 3
 
 image:
-  cassandraRepository: gcr.io/cassandra-operator/cassandra-dev
+  cassandraRepository: gcr.io/cassandra-operator/cassandra
   sidecarRepository: gcr.io/cassandra-operator/cassandra-sidecar-dev
```
zegelin: @benbromhead not sure if we should be removing `-dev` from the sidecar image repo name too?

benbromhead: we probably should.

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java
cin: 👍 I'll move this back into the original method.

helm/cassandra/values.yaml
cin: sorry, i couldn't find the dev image, so i switched this for testing. I eventually made my own images. i'll revert.

PR #146, opened 2019-04-15
Increase Memory limits for Cassandra operator

### LINKED ISSUE #145: Operator resource limits are not enough
The operator limits are defined with low memory. I needed to increase to make it run in my cluster.

I added 300Mi to make operator run without get OOM killed.

rimolive:
When running Cassandra Operator, I get lots of OOM killed events in OpenShift. Because of this, memory limits for operator is changed to 300Mi.

Fixes #145

### DIFF
--- a/examples/common/bundle.yaml
+++ b/examples/common/bundle.yaml
@@ -23,7 +23,7 @@ spec:
         resources:
           limits:
             cpu: 200m
-            memory: 100Mi
+            memory: 300Mi
           requests:
             cpu: 100m
             memory: 50Mi
--- a/examples/common/rbac-bundle.yaml
+++ b/examples/common/rbac-bundle.yaml
@@ -99,7 +99,7 @@ spec:
         resources:
           limits:
             cpu: 200m
-            memory: 100Mi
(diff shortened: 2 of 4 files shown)


rimolive:
@benbromhead Would you mind pointing me where should I change?

benbromhead:
The `values.yaml` file gets templated into the template at https://github.com/instaclustr/cassandra-operator/blob/8c82d16c0a5d47ced8aaa0c9fe8b1e2a701d6492/helm/cassandra/templates/cassandra.yaml#L24

You can find the right values at https://github.com/instaclustr/cassandra-operator/blob/8c82d16c0a5d47ced8aaa0c9fe8b1e2a701d6492/helm/cassandra/values.yaml#L18-L28

DBarthe:
>  maybe change them so they are inline with the resource defaults in th

The default 512Mi in helm for cassandra node also causes OOM (on my workstation at least). I always put 1024Mi...

smiklosovic:
This wont likely be an issue once we switch to Go-based operator. I ll close it once that transition happens.

PR #148, opened 2019-04-15
Adds paging support for various resource lists. 

zegelin:
- Fixes issue where some resources wouldn't be cleaned up correctly because the operator didn't paginate through all the results.


java/operator/src/main/java/com/instaclustr/cassandra/operator/k8s/K8sResourceUtils.java
```diff
@@ -142,19 +146,160 @@ public void deletePersistentVolumeAndPersistentVolumeClaim(final V1Pod pod) thro
...
+            }
+
+            @Override
+            public Collection<V1Service> items() {
+                return serviceList.getItems();
+            }
+
+            @Override
+            public ResourceListIterable.Page<V1Service> nextPage() throws ApiException {
+                final String continueToken = serviceList.getMetadata().getContinue();
```
benbromhead: I can't believe this isn't done by the driver...

PR #150, opened 2019-04-16
Add conditional in Makefiles in order to use other image builders

### LINKED ISSUE #147: Use other build commands for Images
I am building operator code on Fedora, and as Docker in Fedora is another distribution and there is another builder that is OCI compliant, there should be a way to make builds check for this other command and run accordingly.

rimolive:
My Fedora uses [Buildah](https://buildah.io/) to build images with parameters. Because of that, I decided to add some conditionals to look for buildah in the path and run it instead of docker if present.

Fixes #147

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -1,0 +1,1 @@
+BUILD_COMMAND ?= "docker build"
--- a/docker/base-openjre/Makefile
+++ b/docker/base-openjre/Makefile
@@ -1,10 +1,12 @@
+include ../../Makefile.inc
+
 OPENJRE_VERSION := "8u212-b01-1~deb9u1"
 OPENJRE_IMAGE_TAG := stretch-$(subst ~,-,$(OPENJRE_VERSION))
 OPENJRE_IMAGE := $(DOCKER_REPO)base-openjre:$(OPENJRE_IMAGE_TAG)
 
 .PHONY: base-openjre
 base-openjre:
-	docker build \
+	$(BUILD_COMMAND) \
 		--build-arg openjre_version=$(OPENJRE_VERSION) \
 		-t $(DOCKER_REPO)base-openjre \
 		-t $(OPENJRE_IMAGE) \
--- a/docker/cassandra-operator/Makefile
+++ b/docker/cassandra-operator/Makefile
@@ -13,7 +13,7 @@ $(CASSANDRA_OPERATOR_JAR):
 	
 .PHONY: cassandra-operator
 cassandra-operator: $(CASSANDRA_OPERATOR_JAR)
-	docker build \
+	$(BUILD_COMMAND) \
 		--build-arg cassandra_operator_jar=$(CASSANDRA_OPERATOR_JAR) \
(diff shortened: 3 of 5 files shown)


zegelin:
I think having a `BUILD_COMMAND` variable is the right approach.

I suggest that we create a global include file `Makefile.inc` situated in the root directory that's then included by every Makefile. This file can define the variable with a default.

What I don't agree with is including logic to determine if a distro specific alternative command is available (in this case, `buildah`). Instead, that should be left up to the person building the project to override -- eg, via `make BUILD_COMMAND=buildah`

This is similar to how most (simple) Makefiles define variables for the compiler toolchain tools they use (`CC`, `CXX`, `LD`, `CFLAGS`, etc) with the defacto default. A user may override them to use an alternate compiler/toolchain programs or arguments.

rimolive:
Hi @zegelin, I understand what you say. But look at the BUILD_COMMAND variable and you will see that  it does not include only the command name but the build subcommand too. Using BUILD_COMMAND as you described will require the developer how to build using that specific command. In case of buildah, we use `buildah build-using-dockerfile`, different than just `docker build` command.

If you think the developer should know how to build with their specific command, then I will make the change.

docker/base-openjre/Makefile
```diff
@@ -2,9 +2,11 @@ OPENJRE_VERSION := "8u181-b13-2~deb9u1"
 OPENJRE_IMAGE_TAG := stretch-$(subst ~,-,$(OPENJRE_VERSION))
 OPENJRE_IMAGE := $(DOCKER_REPO)/base-openjre:$(OPENJRE_IMAGE_TAG)
 
+BUILD_COMMAND := $(if $(shell which buildah),buildah build-using-dockerfile,docker build)
```
benbromhead: If we could just have build command default to `docker build` and leave the buildah components out of it. You should still be able to over ride it via BUILD_COMMAND

I think that is what @zegelin was getting at

zegelin: Yes exactly. `BUILD_COMMAND` wouldn't just be the name of the binary to run (`docker`, `build`, etc) but can also contain any arguments required (i.e `docker build`, `buildah bud`).

So I propose:
- A global include Makefile in the root directory that defines `BUILD_COMMAND` and sets it to `docker build`.
- For alternate build commands, just invoke `make BUILD_COMMAND='builda bud'`

benbromhead:
Hi @rimolive 

Any updates?

rimolive:
I'm working on it. will send an update soon

benbromhead:
cc @zegelin

Makefile.inc
```diff
@@ -0,0 +1,3 @@
+ifndef BUILD_COMMAND
```
zegelin: Minor nit: The `ifndef` is redundant as make already supports conditionally setting variables with the `?=` operator.
See https://www.gnu.org/software/make/manual/make.html#Setting

rimolive:
I made the change in Makefile.inc

PR #153, opened 2019-04-20
support for ssl using user configmap and secret

DBarthe:
Hello,

This PR documents how to use the operator to deploy a datacenter with ssl encryption (see doc/spec_examples.md).

It also fix a few problems in the helm charts :
* add `userSecretVolumeSource`
* template indentation issue 

It disable openapi validation on `userConfigMapVolumeSource` because of a known bug : https://github.com/kubernetes/kubernetes/pull/76124

### DIFF
--- a/doc/spec_examples.md
+++ b/doc/spec_examples.md
@@ -1,3 +1,103 @@
 ## Specific Examples
 ### TODO - Feel free to submit some examples :)
 For the moment check out the master readme.md
+
+
+### SSL encryption
+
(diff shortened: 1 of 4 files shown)


benbromhead:
Awesome work! Thanks for the contribution and the documentation!!

I'll go ahead and merge this one as it LGTM

I'll also create a seperate Issue to track maybe toggling exposed SSL vs non-ssl ports based on config. 

See #154

PR #155, opened 2019-04-23
Update Cassandra image to include LDAP and Kerberos plugins.

zegelin:
Also enables repository-less Docker builds (i.e., on a local machine)

### DIFF
--- a/docker/cassandra/install-cassandra
+++ b/docker/cassandra/install-cassandra
@@ -1,6 +1,26 @@
 #!/bin/bash -xue
 
+echoerr() { >&2 echo $@; }
+
 cassandra_version=$1
+
+case $cassandra_version in
+3.*)
+    # supported
+    ;;
+
+2.*)
+    # 2.2 pulls in openjre, which pulls in gtk, x11, etc.
+    # also, the default-config is specific to 3.11.x.
+    # 2.2 fails to start due to unrecognised YAML properties.
+
+    # no k8s addons support, yet.
+    ;&
+
+*)
+    echoerr "WARNING: Cassandra version ${cassandra_version} is unsupported."
+    exit 1;
+esac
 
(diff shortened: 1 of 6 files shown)


docker/cassandra/install-cassandra
```diff
@@ -42,6 +62,40 @@ mkdir /etc/cassandra/logback.xml.d
...
+
+    # TODO: switch remaining Maven Central URLs to GitHub.
+    case ${cassandra_version} in
+    3.11.*)
+        curl -SLO "https://github.com/instaclustr/cassandra-ldap/releases/download/3.11.4.4/cassandra-ldap-3.11.4.jar"
+        ;;
+    3.0.*)
+        curl -SLO "https://search.maven.org/remotecontent?filepath=com/instaclustr/cassandra-ldap/3.0.18.0/cassandra-ldap-3.0.18.0.jar"
+        ;;
+    2.2.*)
```
benbromhead: Is it worth doing 2.2? Given the rest of the operator doesnt support it?

docker/cassandra/install-cassandra
```diff
@@ -1,7 +1,27 @@
...
+echoerr() { >&2 echo $@; }
+
 cassandra_version=$1
 
+case $cassandra_version in
+3.*)
+    # supported
+    ;;
+
+2.*)
```
benbromhead: we should probably just remove this one

docker/cassandra/install-cassandra
zegelin: Added this before I tested 2.2. Then I realized that does some weird things on install and even fails to start, hence the test and comments above (line 12+).

Happy to remove this. Your call.

docker/cassandra/install-cassandra
zegelin: Put this here as more of a "here be dragons" to me or anyone else in the future.

PR #158, opened 2019-05-01
fixed building of images

smiklosovic:


docker/base-openjre/Makefile
```diff
@@ -1,4 +1,4 @@
-OPENJRE_VERSION := "8u181-b13-2~deb9u1"
+OPENJRE_VERSION := "8u212-b01-1~deb9u1"
```
smiklosovic: old version is not existing in stretch repo anymore

docker/cassandra-operator/Dockerfile.template
```diff
@@ -1,6 +1,5 @@
-ARG openjre_image
-
-FROM ${openjre_image}
+# please execute 'make' in this directory to build this image
+FROM {openjre_image}
```
smiklosovic: I cant build this image, maybe I have just different  docker  version but according to spec / docs FROM should be the very first command otherwise it  fails

zegelin: This is supported by newer Docker versions: https://docs.docker.com/engine/reference/builder/#understand-how-arg-and-from-interact

smiklosovic:
for reference, the problem is with different versions of Docker we run, I am using

    Client:
    Version:         1.13.1
    API version:     1.26
    Package version: docker-1.13.1-66.git1185cfd.fc29.x86_64
    Go version:      go1.11.5
    Git commit:      1185cfd/1.13.1
    Built:           Mon Mar 18 13:59:32 2019
    OS/Arch:         linux/amd64

    Server:
    Version:         1.13.1
    API version:     1.26 (minimum version 1.12)
    Package version: docker-1.13.1-66.git1185cfd.fc29.x86_64
    Go version:      go1.11.5
    Git commit:      1185cfd/1.13.1
    Built:           Mon Mar 18 13:59:32 2019
    OS/Arch:         linux/amd64
    Experimental:    false

On the other hand, Adam is using

    $ docker version
    Client: Docker Engine - Community
     Version:           18.09.5
     API version:       1.39
     Go version:        go1.12.3
     Git commit:        e8ff056
     Built:             Thu Apr 11 08:08:04 2019
     OS/Arch:           darwin/amd64
     Experimental:      false

It seems that community edition enables this feature of having ARG before FROM but docker in Fedora 29 does not have this feature.

Possible options:
- enforcing to use CE edition of Docker
- merge this or figure out other templating mechanism
- using static name of image

benbromhead:
@smiklosovic you are just on an old version of docker. Find the latest via fedora packages or install individually. 

It's pretty reasonable to expect folk build images using the latest (or a much later) version of docker. The end docker image format should still be backwards compatible.

benbromhead:
If there are any reasons why we shouldn't support the existing implementation let me know, otherwise I'll close this PR

PR #159, opened 2019-05-01
update of openjre to 8u212-b01

smiklosovic:
8u181-b13 is not in Debian anymore 

https://packages.debian.org/stretch/openjdk-8-jre


benbromhead:
Lgtm

PR #162, opened 2019-05-13
initial Sidecar client impl

smiklosovic:
backup is not done yet

### DIFF
--- a/pkg/sidecar/client.go
+++ b/pkg/sidecar/client.go
@@ -1,0 +1,196 @@
+package sidecar
+
+import (
+	"fmt"
+	"github.com/go-resty/resty"
+)
+
+const (
+	ACCEPT                         = "Accept"
+	ApplicationJson                = "application/json"
+	EndpointOperationsDecommission = "operations/decommission"
+	EndpointStatus                 = "status"
+)
+
+////////////
+//// CLIENT
+////////////
+
+type Client struct {
+	Host        string
+	Options     *ClientOptions
+	restyClient *resty.Client
+}
+
+type ClientOptions struct {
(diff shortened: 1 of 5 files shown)


pkg/sidecar/client.go
```diff
@@ -0,0 +1,196 @@
...
+	return &BackupResponse{}, nil
+}
+
+/////////////////
+//// OPERATIONS
+/////////////////
+
+type DecommissionError struct {
+	status   int
+	msg      string
```
alourie: I say let's keep the error itself here, not just the string

pkg/sidecar/client.go
```diff
@@ -0,0 +1,196 @@
...
+	GetStatus() (*StatusResponse, error)
+}
+
+func (client *Client) GetStatus() (*StatusResponse, error) {
+
+	statusResponse := &StatusResponse{}
+
+	response, err := client.restyClient.R().
+		SetHeader(ACCEPT, ApplicationJson).
+		SetResult(&statusResponse).
```
alourie: same here, no need to use &

pkg/sidecar/client_test.go
```diff
@@ -0,0 +1,57 @@
...
+	} else {
+		return host
+	}
+}
+
+func TestSidecarClient_GetStatus(t *testing.T) {
+
+	client := NewSidecarClient(getHost(), &ClientOptions{Secure: false})
+
+	response, e := client.GetStatus()
```
alourie: e -> err :-)

benbromhead:
lgtm cc @zegelin

PR #163, opened 2019-05-14
More Java -> Go porting work. Clusters now start!

zegelin:


deploy/crds/cassandraoperator_v1alpha1_cassandradatacenter_crd.yaml
```diff
@@ -32,7 +32,9 @@ spec:
             cassandraImage:
```
zegelin: Ignore this file -- it's auto-generated.

docker/cassandra/entry-point
```diff
@@ -12,4 +12,4 @@ do
 done
 )
 
-/usr/sbin/cassandra
\ No newline at end of file
+exec /usr/sbin/cassandra
```
zegelin: `exec` to skip child processes and so exit works properly.

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
```diff
@@ -5,16 +5,17 @@ import (
...
-	Cluster v1.LocalObjectReference `json:"cluster"`
-	Replicas int32 `json:"replicas"`
-	CassandraImage string `json:"cassandraImage"`
-	SidecarImage string `json:"sidecarImage"`
-	ImagePullPolicy v1.PullPolicy `json:"imagePullPolicy"`
-	ImagePullSecrets []v1.LocalObjectReference `json:"imagePullSecrets"`
+	// Cluster is either a string or v1.LocalObjectReference
+	//Cluster interface{} `json:"cluster,omitempty"`
+	Cluster          string                    `json:"cluster,omitempty"`
+	Nodes            int32                     `json:"nodes"`
```
zegelin: Changed `replicas` -> `nodes`, to fit within Cassandra parlance.

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
```diff
@@ -5,16 +5,17 @@ import (
...
 // +k8s:openapi-gen=true
 type CassandraDataCenterSpec struct {
-	Cluster v1.LocalObjectReference `json:"cluster"`
-	Replicas int32 `json:"replicas"`
-	CassandraImage string `json:"cassandraImage"`
-	SidecarImage string `json:"sidecarImage"`
-	ImagePullPolicy v1.PullPolicy `json:"imagePullPolicy"`
-	ImagePullSecrets []v1.LocalObjectReference `json:"imagePullSecrets"`
+	// Cluster is either a string or v1.LocalObjectReference
+	//Cluster interface{} `json:"cluster,omitempty"`
```
zegelin: This is "TODO"

pkg/controller/cassandradatacenter/controller.go
```diff
@@ -101,6 +104,13 @@ func (r *CassandraDataCenterReconciler) Reconcile(request reconcile.Request) (re
 		return reconcile.Result{}, err
 	}
 
+	// get (or create a new, unique) Cluster for this CDC
+	cluster := &cassandraoperatorv1alpha1.CassandraCluster{}
+	_ = cluster
```
zegelin: TODO

pkg/controller/cassandradatacenter/controller.go
```diff
@@ -121,14 +134,100 @@ func (r *CassandraDataCenterReconciler) Reconcile(request reconcile.Request) (re
 	// TODO:
 	_, _, _ = nodesService, seedNodesService, statefulSet
 
-
 	return reconcile.Result{}, nil
 }
 
+//func (r *CassandraDataCenterReconciler) getCluster(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, cluster *cassandraoperatorv1alpha1.CassandraCluster) error {
```
zegelin: Ignore all this -- this is all for the Cluster object machinery.

docker/cassandra/entry-point
benbromhead: Just so I'm clear, what was wrong previously?

examples/go/example-datacenter.yaml
```diff
@@ -1,17 +1,15 @@
-
 apiVersion: cassandraoperator.instaclustr.com/v1alpha1
```
benbromhead: Why do we have a separate go folder for these examples? Just overwrite the existing ones

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
benbromhead: Does replicas have a special meaning in K8s tho? E.g. see https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#scale-subresource

Or is this just an internal struct where the name doesn't matter a ton

pkg/apis/cassandraoperator/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -259,7 +259,6 @@ func (in *CassandraDataCenterList) DeepCopyObject() runtime.Object {
 // DeepCopyInto is an autogenerated deepcopy function, copying the receiver, writing into out. in must be non-nil.
 func (in *CassandraDataCenterSpec) DeepCopyInto(out *CassandraDataCenterSpec) {
 	*out = *in
-	out.Cluster = in.Cluster
```
benbromhead: any reason why this line is removed from the autogen code?

pkg/apis/cassandraoperator/v1alpha1/zz_generated.openapi.go
```diff
@@ -210,10 +210,12 @@ func schema_pkg_apis_cassandraoperator_v1alpha1_CassandraDataCenterSpec(ref comm
...
 					"cluster": {
 						SchemaProps: spec.SchemaProps{
-							Ref: ref("k8s.io/api/core/v1.LocalObjectReference"),
+							Description: "Cluster is either a string or v1.LocalObjectReference Cluster interface{} `json:\"cluster,omitempty\"`",
+							Type:        []string{"string"},
+							Format:      "",
 						},
 					},
-					"replicas": {
+					"nodes": {
```
benbromhead: Will we need to keep renaming this / will changing the autogen stuff break things down the line?

pkg/controller/cassandradatacenter/controller.go
```diff
@@ -161,83 +260,77 @@ func (r *CassandraDataCenterReconciler) createOrUpdateNodesService(cdc *cassandr
...
 }
 
 func (r *CassandraDataCenterReconciler) createOrUpdateSeedNodesService(cdc *cassandraoperatorv1alpha1.CassandraDataCenter) (*corev1.Service, error) {
-	seedNodesService := &corev1.Service{ObjectMeta: dataCenterResourceMetadata(cdc, "seeds"),}
+	seedNodesService := &corev1.Service{ObjectMeta: dataCenterResourceMetadata(cdc, "seeds")}
 
 	_, err := controllerutil.CreateOrUpdate(context.TODO(), r.client, seedNodesService, func(_ runtime.Object) error {
 		seedNodesService.Spec = corev1.ServiceSpec{
 			ClusterIP: "None",
 			Ports: []corev1.ServicePort{
```
benbromhead: 7001 for secure internode

pkg/controller/cassandradatacenter/controller.go
```diff
@@ -365,11 +502,151 @@ func (r *CassandraDataCenterReconciler) createOrUpdateOperatorConfigMap(cdc *cas
 
 	volumeMount := &VolumeMount{
```
benbromhead: TODO: add secret volume iirc (both sidecar and C*)

pkg/controller/cassandradatacenter/controller.go
```diff
@@ -365,11 +502,151 @@ func (r *CassandraDataCenterReconciler) createOrUpdateOperatorConfigMap(cdc *cas
...
+	writeProperty("rack", "rack1")
+	writeProperty("prefer_local", "true")
+
+	configMapVolumeAddTextFile(configMap, volumeSource, "cassandra-rackdc.properties", writer.String())
+
+	return nil
+}
+
+// is this seriously not built-in?
+func min(a int64, b int64) int64 {
```
benbromhead: Move to utility class?

docker/cassandra/entry-point
zegelin: `cassandra` becomes a child of `bash`, which is unnecessary, and IIRC can cause problems with things like `SIGTERM` handling. In general, unless the parent needs to be kept around for some reason, `exec` is better.

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
zegelin: Yet to play around with the `/scale` endpoint stuff, but my general understanding is that `SpecReplicasPath` and `StatusReplicasPath` can point to any field -- it doesn't need to be called `Replicas`.

i.e, instead of `specReplicasPath: .spec.replicas` we could have `specReplicasPath: .spec.nodes`.

examples/go/example-datacenter.yaml
zegelin: Yep, fair point.

pkg/apis/cassandraoperator/v1alpha1/zz_generated.deepcopy.go
zegelin: `cluster` went from being a `v1.LocalObjectReference` to ` string`, which I guess (no idea how the deep copy stuff works), doesn't require extra handling or something.

I'm still mucking around with how to best represent clusters and multi-DC -- this is a symptom of that.

pkg/controller/cassandradatacenter/controller.go
zegelin: Port numbers in services don't really matter when we're using it for Pod discovery -- all Pods can talk to any port. But if you don't specify any ports, the service doesn't work. There was a bug ID, but I've lost it..

```// a port needs to be defined for the service to be resolvable (#there-was-a-bug-ID-and-now-I-cant-find-it)
                        .ports(ImmutableList.of(new V1ServicePort().name("internode").port(7000)))```

alourie:
lgtm

PR #167, opened 2019-05-15
splitting of operator

zegelin:


pkg/controller/cassandradatacenter/controller/controller.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// Add creates a new CassandraDataCenter Controller and adds it to the Manager. The Manager will set fields on the Controller
+// and Start it when the Manager is Started.
+func Add(mgr manager.Manager) error {
+	return add(mgr, newReconciler(mgr))
+}
+
+// newReconciler returns a new reconcile.Reconciler
+func newReconciler(mgr manager.Manager) reconcile.Reconciler {
```
zegelin: I don't think this needs to be changed from the original.

pkg/controller/cassandradatacenter/controller/jvm.go
```diff
@@ -0,0 +1,113 @@
+package controller
+
+import (
+	cassandraoperatorv1alpha1 "github.com/instaclustr/cassandra-operator/pkg/apis/cassandraoperator/v1alpha1"
+	"strconv"
+	"strings"
+)
+
+func cassandraJVMOptions(cdc *cassandraoperatorv1alpha1.CassandraDataCenter) (*string, error) {
```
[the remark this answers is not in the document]
zegelin: We don't need to check for errors -- we are writing to an in-memory buffer, not to a file. The Go docs even ignore the result of `Fprintf` in the case of it writing to a `strings.Builder`.

pkg/controller/cassandradatacenter/controller/constants.go
```diff
@@ -0,0 +1,37 @@
+package controller
```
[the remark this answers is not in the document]
smiklosovic: @zegelin another option is to use them in their respective files where they are used and there would be like constants, I dont like the fact they are not clearly visible by mere looking "somewhere"

pkg/controller/cassandradatacenter/controller/jvm.go
```diff
@@ -0,0 +1,113 @@
...
+	// TODO: should this be Limits or Requests?
+	jvmHeapSize := jvmHeapSize(memoryLimit(cdc))
```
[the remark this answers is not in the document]
zegelin: what rabbit hole? This function is just returning the value of a field...

In the original, the logic was all in one place -- now it's split between 4 functions!

pkg/controller/cassandradatacenter/controller/jvm.go
```diff
@@ -0,0 +1,113 @@
...
+
+		if err := writePropertiesConditionally(&writer, []conditionalProperty{
+			{property: "-XX:+UseG1GC",},
+			{property: "-XX:G1RSetUpdatingPauseTimePercent=5",},
+			{property: "-XX:MaxGCPauseMillis=500",},
+			{
+				condition: func() bool {
+					return jvmHeapSize > 16*GIBIBYTE
+				},
+				property: "-XX:InitiatingHeapOccupancyPercent=70",
```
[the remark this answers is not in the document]
zegelin: Again, it's safe to not check for errors in this case.

pkg/controller/cassandradatacenter/controller/jvm.go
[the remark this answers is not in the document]
smiklosovic: thats good news, we can further get rid of that then, I am not sure how it will play with some go code quality tools we can apply here in the future, they would maybe mark them as false positives that we have not checked them and code quality rating could suffer

pkg/controller/cassandradatacenter/controller/jvm.go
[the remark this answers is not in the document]
smiklosovic: the whole resource / limit stuff, we can discuss this privately, whats wrong about four functions? every one just does what it should ...

pkg/controller/cassandradatacenter/controller/configmap.go
```diff
@@ -0,0 +1,194 @@
...
+		return err
+	}
+
+	properties := map[string]string{
+		"dc":           config.cdc.Name,
+		"rack":         "rack1",
+		"prefer_local": "true",
+	}
+
+	if err := writeKeyValueProperties(&writer, properties); err != nil {
```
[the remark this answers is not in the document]
zegelin: Unless we require the functionality elsewhere, I think splitting it up adds complexity. We don't write `.properties` files anywhere else (yet!), and if and when we do, we should look at implementing a compliant writer. The Java code leverages the built-in `Properties.store(...)` function, which correctly handles things like new lines and keys with spaces, unicode escaping, etc.

In this case, we need to write 4 lines to a single in-memory buffer, which is simple enough for a single function.

And again, we don't need to handle errors.

pkg/controller/cassandradatacenter/controller/constants.go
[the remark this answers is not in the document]
zegelin: When, if ever, does `NewLine` ever need to be anything other than `\n`?  We're supporting Unix-compatible only, so windows or Mac (classic!) style new-lines aren't _ever_ going to be a thing.

Various strings, such as the file names, container names, volume mount-points and volume names, are simply an implementation detail and don't need to be constants. Why do these need to be defined somewhere other than the single place they're used?

pkg/controller/cassandradatacenter/controller/jvm.go
```diff
@@ -0,0 +1,113 @@
...
+		"-Xmx%s\n": strconv.FormatInt(jvmHeapSize, 10),
+	}); err != nil {
+		return nil, err
+	}
+
+	// copied from stock jvm.options
+	if !isUsingGC1(jvmHeapSize) {
+
+		// young gen size
+		if err := writeFormattedProperty(&writer, "-Xmn%s\n", strconv.FormatInt(youngGen(jvmHeapSize), 10)); err != nil {
```
zegelin: There is also inconsistency here now too. Instead of everything just using `fprintf`/`fprintln`, we now have 3 functions: `writeFormattedProperty`, `writePropertiesWithNewLine`, and `writePropertiesConditionally`.

Why doesn't `writeFormattedProperty` add a new line and instead still uses `\n` rather than the `NewLine` constant? Does `writePropertiesConditionally` add new lines? Why doesn't everything just use `writePropertiesConditionally`, since the condition is apparently optional?

This is far, far, far more complex than the original, which was simply formatting some stings into a buffer, with a few `if`s thrown in if things were conditional!

pkg/controller/cassandradatacenter/controller/jvm.go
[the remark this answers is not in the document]
zegelin: Sorry, it wasn't clear that this is all very temporary -- there are a few vague comments in the Java code, but I didn't copy them over when re-writing it.

This is all based on the calculations performed in the stock `cassandra-env.sh`. At some-point we need to determine what the best options are and re-write this to have the correct approach.

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -0,0 +1,185 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func CreateOrUpdateStatefulSet(reconciler *CassandraDataCenterReconciler, cdc *cassandraoperatorv1alpha1.CassandraDataCenter, volumeMounts VolumeMounts) (*v1beta2.StatefulSet, error) {
+
+	dataVolumeClaim := getDataVolumeClaim(cdc.Spec.DataVolumeClaimSpec)
```
zegelin: I don't particularly like the name `get` as it implies fetching the object from somewhere, instead of what it's actually doing which is instantiating an object. Perhaps `new` would be a better name?

pkg/controller/cassandradatacenter/controller/constants.go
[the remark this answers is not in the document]
smiklosovic: thats legit, as i said we can move them to their files and get rid of this

pkg/controller/cassandradatacenter/controller/reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+)
+
+var log = logf.Log.WithName("CassandraDataCenterReconciler")
+
+// CassandraDataCenterReconciler reconciles a CassandraDataCenter object
+type CassandraDataCenterReconciler struct {
+	// This client is a split client that reads objects from the cache and writes to the apiserver
+	client client.Client
```
alourie: Can we keep a pointer here too?

smiklosovic: maybe, I am not sure if that was not generated by operator-sdk, I have to check

pkg/controller/cassandradatacenter/controller/operator.go
```diff
@@ -0,0 +1,31 @@
...
+
+	CreateOrUpdateSeedNodesService(cdc *cassandraoperatorv1alpha1.CassandraDataCenter) (*corev1.Service, error)
+
+	CreateOrUpdateOperatorConfigMap(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, seedNodesService *corev1.Service) (*VolumeMount, error)
+
+	CreateOrUpdateStatefulSet(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, volumeMounts VolumeMounts) (*v1beta2.StatefulSet, error)
+
+	Reconcile(request reconcile.Request) (reconcile.Result, error)
+
+	GetClient() client.Client
```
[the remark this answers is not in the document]
smiklosovic: i ve deleted them, these fields in struct, even private, are visible everywhere as it is in one package

pkg/controller/cassandradatacenter/controller/reconciler.go
smiklosovic: i just leave it as it is for now

pkg/controller/cassandradatacenter/controller/reconciler.go
```diff
@@ -0,0 +1,107 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	statefulSet, err := reconciler.CreateOrUpdateStatefulSet(cdc, VolumeMounts{configMapVolumeMount})
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// TODO:
```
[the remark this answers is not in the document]
smiklosovic: left for now

pkg/controller/cassandradatacenter/controller/statefulset.go
smiklosovic: prefixed with `new` instead of `get`

pkg/controller/cassandradatacenter/controller/utils.go
```diff
@@ -0,0 +1,111 @@
+package controller
```
smiklosovic: this whole file is useless so deleted

pkg/controller/cassandradatacenter/controller/constants.go
```diff
@@ -0,0 +1,10 @@
...
+
+// Constants here are used in more than one go file.
+
+const (
+	CassandraInternodePort  = 7000
+	CassandraCqlPort        = 9042
```
zegelin: Minor nit: both CQL and JMX are acronyms, yet one is defined as `Cql` the other `JMX`. It's either `CQL`/`JMX` or `Cql`/`Jmx`.

smiklosovic: sure, will fix

pkg/controller/cassandradatacenter/controller/datacenter.go
```diff
@@ -0,0 +1,24 @@
+package controller
```
zegelin: lets rename this to `metadata`, since its responsible for generating object metatdata values

pkg/controller/cassandradatacenter/controller/configmap.go
```diff
@@ -0,0 +1,190 @@
...
+	cassandraoperatorv1alpha1 "github.com/instaclustr/cassandra-operator/pkg/apis/cassandraoperator/v1alpha1"
+	"gopkg.in/yaml.v2"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"regexp"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"strings"
+)
+
+const (
```
zegelin: I'm still not convinced, sorry. There are numerous other string literals, in various places, that we haven't promoted to constants: the controller name, the container names, the StatefulSet service name, label keys & values, port names, etc.

Why these ones specifically?

Right now, I think I'd prefer that these get re-inlined. We can promote values to constants in the future if and when their value is used in multiple places.

smiklosovic: I have explained that multiple times, its not for having them exported so they can be used multiple times, _only_, it is also for me having a brief look where are thing _without the necessity to dig in the code_.

pkg/controller/cassandradatacenter/controller/controller.go
```diff
@@ -0,0 +1,60 @@
+package controller
```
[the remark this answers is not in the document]
zegelin: I thin we can safely refactor the code scaffolded by the SDK, as long as we're aware of what's going on. It leverages various `init()` functions to "register" the various controllers and CRD types. IMO, this is just extra fluff and we can rewrite it as we see fit.

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -0,0 +1,185 @@
...
+			{Name: "http", ContainerPort: 4567},
+		},
+		VolumeMounts: []corev1.VolumeMount{
+			{Name: dataVolumeClaim.Name, MountPath: "/var/lib/cassandra"},
+			{Name: podInfoVolume.Name, MountPath: "/etc/pod-info"},
+		},
+	}
+}
+
+func getPodInfoVolume() *corev1.Volume {
```
zegelin: `new`

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -0,0 +1,185 @@
...
+					{Path: "annotations", FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.annotations"}},
+					{Path: "namespace", FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.namespace"}},
+					{Path: "name", FieldRef: &corev1.ObjectFieldSelector{FieldPath: "metadata.name"}},
+				},
+			},
+		},
+	}
+}
+
+func getDataVolumeClaim(dataVolumeClaimSpec corev1.PersistentVolumeClaimSpec) *corev1.PersistentVolumeClaim {
```
zegelin: `new`

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -0,0 +1,185 @@
...
+	})
+
+	if err != nil {
+		return nil, err
+	}
+
+	return statefulSet, err
+}
+
+func getStatefulSetSpec(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, podSpec *corev1.PodSpec, dataVolumeClaim *corev1.PersistentVolumeClaim) v1beta2.StatefulSetSpec {
```
zegelin: `new`

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -0,0 +1,185 @@
...
+		Selector:    &metav1.LabelSelector{MatchLabels: podLabels},
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{Labels: podLabels},
+			Spec:       *podSpec,
+		},
+		VolumeClaimTemplates: []corev1.PersistentVolumeClaim{*dataVolumeClaim},
+	}
+}
+
+func getPodSpec(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, podInfoVolume *corev1.Volume, containers []corev1.Container) *corev1.PodSpec {
```
zegelin: `new`

pkg/controller/cassandradatacenter/controller/statefulset.go
zegelin: Since this function just returns a struct that is in no-way parameterised, perhaps its better to declare it as a package-local `var`? (Can those be `const`/`static`?)

pkg/controller/cassandradatacenter/controller/operator.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	cassandraoperatorv1alpha1 "github.com/instaclustr/cassandra-operator/pkg/apis/cassandraoperator/v1alpha1"
+	"k8s.io/api/apps/v1beta2"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+type CassandraOperator interface {
+	CreateOrUpdateNodesService(cdc *cassandraoperatorv1alpha1.CassandraDataCenter) (*v1.Service, error)
```
zegelin: Why do these need to be interface functions? It looks like their receiver method implementations just call to the global functions...

pkg/controller/cassandradatacenter/controller/jvm.go
```diff
@@ -0,0 +1,113 @@
...
+
+	// OOM Error handling
+	if err := writePropertiesWithNewLine(&writer, []string{
+		"-XX:+HeapDumpOnOutOfMemoryError",
+		"-XX:+CrashOnOutOfMemoryError",
+	}); err != nil {
+		return nil, err
+	}
+
+	// TODO: maybe tune -Dcassandra.available_processors=number_of_processors - Wait till we build C* for Java 11
```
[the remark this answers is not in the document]
zegelin: Better supported I think. This is a comment from @benbromhead IIRC.

pkg/controller/cassandradatacenter/controller/jvm.go
```diff
@@ -0,0 +1,86 @@
+package controller
```
zegelin: Any reason why this is explicitly split into its own file? The ConfigMap contains everything related to Cassandras config, including JVM settings.

pkg/controller/cassandradatacenter/controller/statefulset.go
smiklosovic: are you sure it wont be parametrised? I would leave it as it is, plus we have other "new" methods so I dont see the point to having it like var just because it does not accept any arguments.

PR #169, opened 2019-05-17
Allow the operator to clean up after itself with pvcs 

benbromhead:
- just flag based for the moment.

### DIFF
--- a/java/operator/src/main/java/com/instaclustr/cassandra/operator/Operator.java
+++ b/java/operator/src/main/java/com/instaclustr/cassandra/operator/Operator.java
@@ -46,7 +46,7 @@ public class Operator implements Callable<Void> {
         String namespace = "default";
 
         @Option(names = {"-p", "--allow-pvc-cleanup"}, description = "")
-        Boolean allowPVCCleanups = false;
+        boolean allowPVCCleanups = false;
     }
 
 //    public static class K8sClientOptions {
--- a/java/operator/src/main/java/com/instaclustr/cassandra/operator/Operator.java
+++ b/java/operator/src/main/java/com/instaclustr/cassandra/operator/Operator.java
@@ -1,7 +1,7 @@
 package com.instaclustr.cassandra.operator;
 
-import ch.qos.logback.classic.Level;
 import com.google.inject.*;
+import com.instaclustr.cassandra.operator.configuration.DeletePVC;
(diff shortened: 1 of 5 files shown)


java/operator/src/main/java/com/instaclustr/cassandra/operator/Operator.java
```diff
@@ -49,6 +44,9 @@
     static class OperatorOptions {
         @Option(names = {"-n", "--namespace"}, description = "")
         String namespace = "default";
+
+        @Option(names = {"-p", "--allow-pvc-cleanup"}, description = "")
+        Boolean allowPVCCleanups = false;
```
zegelin: Minor nit: Probably better to use the primitive `boolean` type rather than the boxed `Boolean` here (and where it's injected). `Boolean` is nullable.

Not a big issue here, but I always see usages of `Boolean` and other boxed types in fields or parameters as code smells. They're really only useful in collections or in the case when the value is "optional" (e.g., can be unset).

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterDeletionController.java
```diff
@@ -24,13 +25,17 @@
 
     private final K8sResourceUtils k8sResourceUtils;
 
+    final Boolean allowCleanups;
```
zegelin: Minor nit: this probably could be `private` too.

java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterDeletionController.java
```diff
@@ -106,6 +96,12 @@ public void deleteDataCenter() throws Exception {
                 }
             });
 
+
+            // delete persistent volumes & persistent volume claims
+            if(allowCleanups) {
```
zegelin: Minor nit: code style is space between `if` and opening bracket: eg: `if (allowCleanups) {`

java/operator/src/main/java/com/instaclustr/cassandra/operator/Operator.java
benbromhead: For some reason I thought that primitives couldn't be injected...

benbromhead:
Addressed nits

PR #171, opened 2019-05-20
scaling up and down

smiklosovic:


pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -183,3 +295,147 @@ func newDataVolumeClaim(dataVolumeClaimSpec corev1.PersistentVolumeClaimSpec) *c
...
+
+	for _, pod := range pods {
+		key := pod
+		podClients[&key] = sidecar.NewSidecarClient(pod.Status.PodIP, &options)
+	}
+
+	return podClients
+}
+
+func cassandraStatuses(podClients map[*corev1.Pod]*sidecar.Client) map[*corev1.Pod]string {
```
smiklosovic: I am not sure if we should not make go routines write to channel instead of directly putting that into map but it does work just fine.

pkg/sidecar/client.go
```diff
@@ -140,19 +156,8 @@ func (e *StatusError) Error() string {
 
 type OperationMode string
```
alourie: Drop this too?

PR #175, opened 2019-05-22
Initial implementation of e2e tests in Go

smiklosovic:


test/e2e/cassandra_test.go
```diff
@@ -0,0 +1,69 @@
...
+	t.Logf("Testing Cassandra cluster will be created in namespace %s", namespace)
+
+	name := "test-dc-cassandra"
+	statefulSetName := "cassandra-test-dc-cassandra"
+
+	cassandraDC := defaultNewCassandraDataCenter(name, namespace)
+	cassandraDC.Spec.Nodes = 3
+
+	if err := f.Client.Create(goctx.TODO(), cassandraDC, cleanupOptions); err != nil {
+		t.Fatalf("Unable to create Cassandra DC object of %d node.", cassandraDC.Spec.Nodes)
```
zegelin: Is it worth including `err` in the printed output too, so that we can see why it failed?

Gopkg.lock
```diff
@@ -349,6 +406,14 @@
   revision = "0ff49de124c6f76f8494e194af75bde0f1a49a29"
   version = "v1.1.6"
 
+[[projects]]
```
zegelin: Not really knowing how dep works, but I'm assuming these are transitive dependencies of things like the test framework. Do they need to be explicitly specified here?

test/e2e/utils.go
```diff
@@ -0,0 +1,161 @@
...
+	}
+
+	t.Logf("statefulset %s in namespace %s is fully available.", name, namespace)
+
+	return nil
+}
+
+func podsSidecars(t *testing.T, f *framework.Framework, namespace string) (map[*v1.Pod]*sidecar.Client, error) {
+
+	pods, err := f.KubeClient.CoreV1().Pods(namespace).List(metav1.ListOptions{})
```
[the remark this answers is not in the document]
smiklosovic: there's a possibility to run it in a randomly named namespace so these pods will be the only pods running there

test/e2e/utils.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+func checkAllNodesInNormalMode(t *testing.T, f *framework.Framework, namespace string) {
+	if clients, err := podsSidecars(t, f, namespace); err != nil {
+		t.Fatalf("Error getting pod's sidecars: %v", err)
+	} else {
+		for pod, client := range clients {
+			if status, err := client.GetStatus(); err != nil {
+				t.Fatalf("Error occured while calling sidecar status endpoint for node %s in namespace %s", pod.Status.PodIP, namespace)
+			} else if status.OperationMode != "NORMAL" {
```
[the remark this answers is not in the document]
smiklosovic: yeah, it was like that, we can move it back

Gopkg.lock
smiklosovic: this was generated automatically by running `dep ensure`, I wouldnt touch this stuff by hand.

test/e2e/utils.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if err != nil {
+		t.Fatalf("Failed to initialize cluster resources: %v", err)
+	}
+
+	t.Log("Initialized cluster resources")
+
+	namespace, err := ctx.GetNamespace()
+
+	if err != nil {
```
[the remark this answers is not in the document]
smiklosovic: I dont want to do that as I need access to namespace and it is not available out of that if statement and I dont want to if that all below.

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -101,7 +96,7 @@ func CreateOrUpdateStatefulSet(
 
 				log.Info("No Cassandra nodes have been decommissioned. Decommissioning the newest one " + newestPod.Name)
 
-				if _, err := sidecar.NewSidecarClient(newestPod.Status.PodIP, &sidecarClientOptions).DecommissionNode(); err != nil {
+				if _, err := sidecar.NewSidecarClient(newestPod.Status.PodIP, &sidecar.DefaultSidecarClientOptions).DecommissionNode(); err != nil {
```
zegelin: Can we re-use the `clients` map above for this?

test/e2e/utils.go
```diff
@@ -0,0 +1,192 @@
...
+		t.Fatalf("Could not obtain namespace: %v", err)
+	}
+
+	f := framework.Global
+
+	if err := e2eutil.WaitForOperatorDeployment(t, f.KubeClient, namespace, "cassandra-operator", 1, retryInterval, timeout); err != nil {
+		t.Fatal("Timeout of Cassandra operator deployment has occurred.", err)
```
zegelin: What's the diff between `t.Fatalf` and `t.Fatal`? ie line 29 vs line 43?

test/e2e/utils.go
```diff
@@ -0,0 +1,192 @@
...
+}
+
+func waitForStatefulset(t *testing.T, f *framework.Framework, namespace string, name string, expectedReplicas int32) {
+
+	err := wait.Poll(retryInterval, timeout, func() (done bool, err error) {
+		statefulSet, err := f.KubeClient.AppsV1().StatefulSets(namespace).Get(name, metav1.GetOptions{})
+
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				t.Logf("Waiting for availability of %s statefulset in namespace %s\n", name, namespace)
```
zegelin: Is the `\n` necessary?

test/e2e/utils.go
```diff
@@ -0,0 +1,192 @@
...
+		if replicas != readyReplicas {
+			t.Logf("Waiting for full availability of %s statefulset in namespace %s - (%d/%d)", name, namespace, readyReplicas, replicas)
+			return false, nil
+		}
+
+		return true, nil
+	})
+
+	if err != nil {
+		t.Fatalf("Timeout for statefulset %s in namespace %s has occurred.", name, namespace)
```
zegelin: Maybe print `err`?

test/e2e/utils.go
smiklosovic: Fatal is equivalent to Log followed by FailNow. Fatalf is equivalent to Logf followed by FailNow. Log is just a logger, with Logf, you can format that message. In this particular case, it should be Fatalf.

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -101,8 +97,12 @@ func CreateOrUpdateStatefulSet(
 
 				log.Info("No Cassandra nodes have been decommissioned. Decommissioning the newest one " + newestPod.Name)
 
-				if _, err := sidecar.NewSidecarClient(newestPod.Status.PodIP, &sidecarClientOptions).DecommissionNode(); err != nil {
-					return errors.New("Failure occurred while decommissioning a Cassandra node upon scaling a cluster down: " + err.Error())
+				if clientForNewestPod := sidecar.ClientForPod(clients, newestPod); clientForNewestPod != nil {
```
zegelin: `clients` is a map, with a key type of `corev1.Pod*`. `newestPod` is also a `corev1.Pod*`.
AFAIK, pointers are comparable, so you should just be able to do `clients[newestPod]` to get the client for the newest pod.

smiklosovic: it doesnt work like that ive tried that approach, i can revisit though

smiklosovic: @zegelin 

`newestPod` here is not a pointer. `allPods` is just an array `[]corev1.Pod` (Kubernetes API returns that). Then we want to know the newest pod so I do `allPods[len(allPods)-1]`. This is not a pointer again. Hence I can not use it as a key to the map.

Now `SidecarClients` returns a map of pointers to pods as keys and pointers to clients as values while it is taking that array of pods from Kubernetes. Check this:

    func SidecarClients(pods []corev1.Pod, clientOptions ClientOptions) map[*corev1.Pod]*Client {

        podClients := make(map[*corev1.Pod]*Client)

        for _, pod := range pods {
            key := pod                      // <----- see this?
            podClients[&key] = NewSidecarClient(pod.Status.PodIP, &clientOptions)
        }

	return podClients
}

Now without `key := pod`, if I was about to put a client like `podClients[pod]`, it would, interestingly, overwrite what I put there, because Go uses same address for `pod` in that loop. Hence I was ending up with just one entry in that map after the loop was done.

Check this where it is explained in depth:

https://stackoverflow.com/questions/48826460/using-pointers-in-a-for-loop-golang

docker/cassandra-operator/Makefile
```diff
@@ -2,7 +2,7 @@
 .PHONY: cassandra-operator
 cassandra-operator:
 	docker build \
-		-t $(DOCKER_REPO)cassandra-operator \
+		-t $(DOCKER_REPO)/cassandra-operator \
```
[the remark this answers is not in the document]
zegelin: Yep. `DOCKER_REPO` should contain the `/`.
Not sure how to enforce this, but some thoughts:
- have a check to ensure that `DOCKER_REPO` is either empty or ends in a `/`.
- write a function which correctly joins `DOCKER_REPO` with the image name, inserting a `/` if necessary.

The last option is probably the best as it would allow `DOCKER_REPO` to be specified with or without a training `/`.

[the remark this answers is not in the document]
benbromhead: yeah I think its becoming clear that from an ergonomics perspective the slash in the env var is annoying. This nit shouldn't block the merge if we revert it, but @zegelin we should just have something that handles it (e.g. your function) so that it doesn't matter and it's easier

benbromhead:
ok lgtm

PR #177, opened 2019-05-30
More refactoring & cleanup

zegelin:
This is a work-in-progress, open for comment.


docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
```diff
@@ -0,0 +1 @@
+ulimit -l unlimited
```
zegelin: Moved this from `ConfigMap.go` to here, based on the TODO.

`default-config/cassandra-env.sh` (aka, `/etc/cassandra/cassandra-env.sh` inside the built image) is blank.
Open for discussion as to whether this line (and similar ones) should go in a _fragment file_ (as this PR introduces) or into the "main" `cassandra-env.sh` file.

The intent of _fragment files_ is to allow for easy additions of new config settings, especially for optional features, by either the operator (e.g, if prometheus support is enabled, drop in a fragment that appends to `JVM_OPTS`) or by users (set some custom JVM params, alter the environment, etc).

Fragments are easy to add, but harder to remove -- the user would have to create a config map containing an empty file with the same name (i.e, `cassandra-env.sh.d/002-cassandra-limits.sh` in this case).

pkg/controller/cassandradatacenter/controller/configmap.go
```diff
@@ -12,48 +12,31 @@ import (
...
-		volumeSource:     volumeSource,
-		seedNodesService: seedNodesService,
-	}
-
-	_, err := controllerutil.CreateOrUpdate(context.TODO(), reconciler.client, config.configMap, func(_ runtime.Object) error {
-
-		if err := addCassandraYamlOverrides(config); err != nil {
-			return err
+	_, err := controllerutil.CreateOrUpdate(context.TODO(), rctx.client, configMap, func(_ runtime.Object) error {
+		addFileFn := func (path string, data string) {
```
zegelin: Each `addXYZ` function needs to modify both the `ConfigMap` and `ConfigMapVolumeSource` objects to add a "file". Previously we were passing around a `configurationResources` "context" object that contained these as fields so that each `addXYZ` function could call `configMapVolumeAddTextFile(...)`, but I feel that becomes a "god struct" -- better to break it up so that each function only takes its direct dependencies.

This IMO is cleaner, but I had two options:
- this approach, with a function parameter/reference/lambda (what's the Go term?) being passed around
- a struct that contains both the `ConfigMap` and `ConfigMapVolumeSource`, with a `addFile()` method receiver

pkg/controller/cassandradatacenter/controller/configmap.go
```diff
@@ -95,34 +74,19 @@ func addCassandraGossipingPropertyFileSnitchProperties(config configurationResou
...
+func addPrometheusSupport(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, addFileFn func (path string, data string)) {
+	if cdc.Spec.PrometheusSupport {
+		addFileFn(
 			"cassandra-env.sh.d/001-cassandra-exporter.sh",
 			"JVM_OPTS=\"${JVM_OPTS} -javaagent:${CASSANDRA_HOME}/agents/cassandra-exporter-agent.jar=@${CASSANDRA_CONF}/cassandra-exporter.conf\"",
 		)
 	}
 }
 
-func addCassandraLimits(config configurationResources) {
```
zegelin: Moved into the image.

pkg/controller/cassandradatacenter/controller/configmap.go
```diff
@@ -157,14 +121,90 @@ func addCassandraYamlOverrides(config configurationResources) error {
...
+		panic(err)
 	}
 
-	configMapVolumeAddTextFile(config.configMap, config.volumeSource, "cassandra.yaml.d/001-operator-overrides.yaml", string(data))
+	addFileFn("cassandra.yaml.d/001-operator-overrides.yaml", string(data))
+}
+
+
+
+func addCassandraJVMOptions(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, addFileFn func (path string, data string)) {
```
zegelin: Moved from `jvm.go`, which seemed like an arbitrary split between files -- everything config related is here, except for JVM config?

pkg/controller/cassandradatacenter/controller/ports.go
```diff
@@ -0,0 +1,18 @@
+package controller
```
zegelin: This currently isn't used, but included for discussion.

Following the discussion on constants, it was determined that any value that's used in more than one place probably should become a constant. The port numbers (and their names) are used in both the container and service definitions, so IMO they should be constants.

Issue is, services hold `ServicePort` structs and containers hold `ContainerPort` structs.

Pending questions:
- do the `ServicePort` and `ContainerPort` names need to match?
- what if we need to define service or container ports that leverage other fields of their respective structs?

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -132,66 +52,44 @@ func CreateOrUpdateStatefulSet(
...
 	}
 }
 
-func newPodSpec(
-	cdc *cassandraoperatorv1alpha1.CassandraDataCenter,
-	podInfoVolume *corev1.Volume,
-	containers []corev1.Container,
-) *corev1.PodSpec {
+func newPodSpec(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, volumes []corev1.Volume, containers []corev1.Container) *corev1.PodSpec {
+	// TODO: any way to pass volumes and containers as pointers?
```
zegelin: As mentioned in this TODO, should we be passing slices-of-pointers for both volumes and containers? AFAIK there isn't an easy (or built-in) way to convert `[]*corev1.Volume` to `[]corev1.Volume` except a for loop.

pkg/controller/cassandradatacenter/controller/statefulset.go
```diff
@@ -289,13 +155,95 @@ func newPodInfoVolume() *corev1.Volume {
...
-func newDataVolumeClaim(dataVolumeClaimSpec corev1.PersistentVolumeClaimSpec) *corev1.PersistentVolumeClaim {
+func newDataVolumeClaim(dataVolumeClaimSpec *corev1.PersistentVolumeClaimSpec) *corev1.PersistentVolumeClaim {
 	return &corev1.PersistentVolumeClaim{
 		ObjectMeta: metav1.ObjectMeta{Name: "data-volume"},
-		Spec:       dataVolumeClaimSpec,
+		Spec:       *dataVolumeClaimSpec,
 	}
 }
 
+//func scaleSafely() {
```
zegelin: TODO

docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
benbromhead: specifically ulimits etc and other bits should be part of the privileged initi container? not part of the C* container

zegelin: In the Java code, inject a fragment into the Cassandra container, identical to this.
But you're right, we're not calling `sysctl` anywhere -- I missed the init container.
Should the ulimits call be in the init container too?

zegelin: See https://github.com/instaclustr/cassandra-operator/blob/master/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java#L412

benbromhead: this is what gets set within the priv container https://github.com/instaclustr/cassandra-operator/blob/8fed62125a785025fe3868ce579b809f9f07cc5e/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java#L338-L345

then the C* container itself has the following kernel capabilities assigned:

https://github.com/instaclustr/cassandra-operator/blob/8fed62125a785025fe3868ce579b809f9f07cc5e/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java#L150-L153

should the actual init container have those kernel capabilities instead and the C* container is less privileged?? I'm not sure, but that would feel to me like a good reduction of permissions...

zegelin: _dives deep into Kernel security contexts_

benbromhead: C* definitely needs IPC_LOCK for itself, so can't move that
ulimit unlimited is the amount it can memlock, so again those two on the C* container are fine as its that process itself that needs those perms

SYS_RESOURCE  http://man7.org/linux/man-pages/man7/capabilities.7.html is file refs 

iirc the init container had to be done as that was a sysctl setting rather than a kernal capability on a process?

pkg/controller/cassandradatacenter/controller/configmap.go
alourie: I personally like this approach. It's quite clean and very common in Go.

Also, I read a bit on receiver functions vs general functions, and it seems that you use receiver function when you either implement an interface, group methods or changing the state of the struct (that is the receiver itself). In other cases you use just functions.

docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
zegelin: Yeah, thinking about it more...

Pods are basically container groups. We need to run sysctl in a privileged container, because that's essentially modifying a host kernel parameter. It will persist even after the container/pod/etc has terminated.

`ulimits` is a bash built-in, and it modifies the limits to all subprocesses started inside that instance of the shell, but not other processes, and will be "reset" (more, forgotten) when the container terminates. Because we want the limits to apply specifically to Cassandra, we run it before we exec the JVM.

benbromhead: ok makes sense

pkg/controller/cassandradatacenter/controller/controller.go
```diff
@@ -26,27 +26,25 @@ func newReconciler(mgr manager.Manager) reconcile.Reconciler {
...
+		OwnerType:    &cassandraoperatorv1alpha1.CassandraDataCenter{},
+	}
 	for _, t := range []runtime.Object{&corev1.Service{}, &v1beta2.StatefulSet{}, &corev1.ConfigMap{}} {
-		err = c.Watch(&source.Kind{Type: t}, &handler.EnqueueRequestForOwner{
-			IsController: true,
-			OwnerType:    &cassandraoperatorv1alpha1.CassandraDataCenter{},
-		})
-		if err != nil {
-			return err
+		if err = c.Watch(&source.Kind{Type: t}, requestForOwnerHandler); err != nil {
```
smiklosovic: interesting, can we really reuse it just like that?

pkg/controller/cassandradatacenter/controller/configmap.go
zegelin: > changing the state of the struct

Which sort-of applies in this case. E.g., say we had a struct `ConfigMapArchive`, that had both the CM and CMVS as fields, then `configMapArchive.addFile(<path>, <data>)` makes sense too, right?

docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
smiklosovic: I would go with fragments unless it is something absolutely crucial for whole container / pod to work correctly which would render it unusable.

pkg/controller/cassandradatacenter/controller/configmap.go
alourie: yes, that would make sense.

I still like this current approach better.

pkg/controller/cassandradatacenter/controller/statefulset.go
zegelin: Actually, I'm just going to remove this function since it basically just wraps the creation of a very simple struct.

pkg/controller/cassandradatacenter/controller/configmap.go
smiklosovic: +1 on narrowing down what we just need to avoid god structs but it requires discipline not to make it god struct again as it might seem tempting to "just pass this little bit into that too" :)

smiklosovic: i would go with this approach as is

smiklosovic: I am just thinking how we would unit test the functions we are passing this addFileFn into. While calling it, I would need to make my own implementation of addFileFn?

zegelin: Yes, but that means its easy to mock out into a no-op, or something that checks the result easily...

pkg/controller/cassandradatacenter/controller/ports.go
smiklosovic: I somehow came up with this ...

```
package controller

import v1 "k8s.io/api/core/v1"

type port struct {
	name string
	port int32
}

type PortFactory struct {
	servicePort   v1.ServicePort
	containerPort v1.ContainerPort
}

var cassandraJmxPort = port{"jmx", 7199}

func (myPort *PortFactory) ContainerPort(p port) v1.ContainerPort {
	myPort.containerPort = v1.ContainerPort{Name: p.name, ContainerPort: p.port}

	// other operations on myPort.containerPort

	return myPort.containerPort
}

func (myPort *PortFactory) ServicePort(p port) v1.ServicePort {

	myPort.servicePort = v1.ServicePort{Name: p.name, Port: p.port}

	// other operations on myPort.servicePort

	return myPort.servicePort
}

/*
	Would be used as

	ctx := portContext{
		name: "",
		port: 1234,
	}

	var servicePort v1.ServicePort = PortFactory.ServicePort(PortFactory{}, ctx)
*/
```

pkg/controller/cassandradatacenter/controller/reconciler.go
```diff
@@ -15,58 +16,66 @@ var log = logf.Log.WithName("CassandraDataCenterReconciler")
 // CassandraDataCenterReconciler reconciles a CassandraDataCenter object
 type CassandraDataCenterReconciler struct {
 	// This client is a split client that reads objects from the cache and writes to the apiserver
-	client client.Client
+	client client.Client // TODO: pointer?
```
smiklosovic: potentially, I think it was generated like that ...

pkg/controller/cassandradatacenter/controller/configmap.go
```diff
@@ -157,14 +121,90 @@ func addCassandraYamlOverrides(config configurationResources) error {
...
+	jvmHeapSize := maxInt64(minInt64(memoryLimit/2, GIBIBYTE), minInt64(memoryLimit/4, 8*GIBIBYTE))
+
+	youngGenSize := youngGen(jvmHeapSize)
+
+	useG1GC := jvmHeapSize > 8*GIBIBYTE
 
-	return nil
+	var writer strings.Builder
+
+	_, _ = fmt.Fprintf(&writer, "-Xms%d\n", jvmHeapSize) // min heap size
```
alourie: Sorry, this looks too weird. Can we just ignore the outputs here? I know we've discussed this before, but I'm pretty sure it would be fine in this case.

Also, strings.Builder has a WriteString method, so maybe we can do something like

`writer.WriteString(....)` or `writer.WriteString(fmt.Printf(...))`, in some sense it's more clear what it does.

pkg/controller/cassandradatacenter/controller/ports.go
smiklosovic: Maybe this is even better:

```
package controller

import v1 "k8s.io/api/core/v1"

type port struct {
	name string
	port int32
}

type PortFactory interface {
	ContainerPort() v1.ContainerPort
	ServicePort() v1.ServicePort
}

var cassandraJmxPort = port{"jmx", 7199}

func (myPort port) ContainerPort() v1.ContainerPort {

	containerPort := v1.ContainerPort{Name: myPort.name, ContainerPort: myPort.port}

	// other operations

	return containerPort
}

func (myPort port) ServicePort() v1.ServicePort {

	servicePort := v1.ServicePort{Name: myPort.name, Port: myPort.port}

	// other operations
	
	return servicePort
}
```

port struct itself would  have these methods on it, from which  you can form whatever specific port you want

pkg/controller/cassandradatacenter/controller/statefulset.go
smiklosovic: I would leave it as is, we are using new... methods in createOrUpdateStatefulSet so why just use direct initialisation there? it is more consistent right now too, if we change this and it starts to be complicated we would move it into method anyway

pkg/controller/cassandradatacenter/controller/controller.go
alourie: yes, we can. `err` is just a var that's already been initialised once.

pkg/controller/cassandradatacenter/controller/reconciler.go
zegelin: Apparently it already is, because all the methods of all the implementations take pointers.

alourie: yea, no need to make it a pointer. `client.Client` is an interface, it would make no sense keeping a pointer to an interface.

pkg/controller/cassandradatacenter/controller/controller.go
smiklosovic: by my concern is about the "semantics" of that, is not that somehow initialised / used differently subsequently? There is on more field in that struct and I am not sure we can just iterate over that and reuse it there ...

pkg/controller/cassandradatacenter/controller/ports.go
alourie: In this case, there's nothing in `PortFactory` that we really need to create the results. In many cases, we'd simply use functions `func ContainerPort()...` and `func ServicePort` to do the same. The receiver in this case doesn't really bring any value. Also, we can do it without the `port` type, just provide `name, port` vars into the creator functions.

We do a similar thing in gocql when creating a new Session object via the `NewSession(options...)` call.

pkg/controller/cassandradatacenter/controller/reconciler.go
smiklosovic: just to be sure we are on the same page, this is not Sidecar client, its some internal client to kubernetes from operator ...

pkg/controller/cassandradatacenter/controller/ports.go
smiklosovic: ah yes, we can omit that interface completely, thats true ... I like that from usage point of view, there is some "generic port" you call functions on to "morph" into whatever you want

PR #193, opened 2019-07-05
implemented cache for operations

smiklosovic:


smiklosovic:
@zegelin forget this, i realised it does not do what it should ...

PR #197, opened 2019-07-09
Implemented filtering of operations

smiklosovic:


java/sidecar-common/src/main/java/com/instaclustr/sidecar/operations/OperationsFilter.java
```diff
@@ -0,0 +1,35 @@
...
+import com.instaclustr.sidecar.validation.Enum;
+
+public class OperationsFilter {
+
+    @JsonSerialize(converter = OperationTypeToStringConverter.class)
+    @JsonDeserialize(converter = StringToOperationTypeConverter.class)
+    public final OperationType type;
+
+    @Enum(enumClass = Operation.State.class, ignoreCase = true)
+    public final String state;
```
smiklosovic: Here it is `string` in order to put Enum constrain on that so we will have fully under control the returning message when user does not specify one of states, like this:

    [
        {
            "message": "Value complet is invalid for enumeration class com.instaclustr.sidecar.operations.Operation$State. Possible values: PENDING,RUNNING,COMPLETED,FAILED",
            "messageTemplate": "Value complet is invalid for enumeration class com.instaclustr.sidecar.operations.Operation$State. Possible values: PENDING,RUNNING,COMPLETED,FAILED",
            "path": "OperationsResource.operations.arg0.state",
            "invalidValue": "complet"
        }
    ]

If it was just `State`, it would say that there is not such enum / it could not be deserialised, but we would not know what values are possible and so on.

java/sidecar-common/src/main/java/com/instaclustr/sidecar/operations/OperationRequest.java
```diff
@@ -14,7 +14,7 @@
 @JsonIgnoreProperties(ignoreUnknown = true)
 public abstract class OperationRequest {
 
-    @JsonProperty("type")
+    @JsonIgnore
```
smiklosovic: no worries, this will be still there because it is "type" info

zegelin: Why does request even have a "type" field, and how does this conflict with the `@JsonTypeInfo` property?

zegelin: The type of a request _is its class type_.

java/sidecar-common/src/main/java/com/instaclustr/sidecar/operations/OperationsResource.java
```diff
@@ -46,6 +46,12 @@ public Operation getOperationById(@NotNull @PathParam("id") final UUID id) {
         return operationsService.operation(id).orElseThrow(NotFoundException::new);
     }
 
+    @POST
```
zegelin: POST requests should modify state. This should be GET request since we're just "getting" a filtered set of operations.

As such, it can probably be merged with the main GET method above.

java/sidecar-common/src/main/java/com/instaclustr/sidecar/operations/OperationsResource.java
```diff
@@ -46,6 +46,12 @@ public Operation getOperationById(@NotNull @PathParam("id") final UUID id) {
...
+    @Path("filter")
```
zegelin: Typically filtering is done via query parameters, rather than a request body.

e.g., `GET /operations?type=backup;type=cleanup`

java/sidecar-common/src/main/java/com/instaclustr/sidecar/operations/OperationsFilter.java
smiklosovic: outdated

java/sidecar-common/src/main/java/com/instaclustr/sidecar/http/JerseyHttpServerModule.java
```diff
@@ -126,13 +120,4 @@ public boolean configure(final FeatureContext context) {
         }
     }
 
-    static class DebugMapper implements ExceptionMapper<Throwable> {
```
smiklosovic: moved to jersey package, seems to be better fit there ...

java/sidecar-common/src/main/java/com/instaclustr/sidecar/jackson/MapBackedTypeIdResolver.java
```diff
@@ -28,29 +29,26 @@ public String idFromValue(final Object value) {
...
-
-        final OperationType operationType = typeMappings.keySet().stream().filter(type -> type.toString().toLowerCase().equals(id.toLowerCase())).findFirst().orElse(null);
+    public Class<? extends T> classFromId(final String id) {
+        return typeMappings.get(id);
+    }
 
-        if (operationType == null) {
-            return null;
-        }
+    public Set<String> typeIds() {
```
smiklosovic: I need this for know what types we have in param converter and its subsequent exception thrown from there

java/sidecar-common/src/main/java/com/instaclustr/sidecar/operations/Operation.java
```diff
@@ -80,4 +88,16 @@ public String toString() {
                 .add("startTime", startTime)
                 .toString();
     }
+
+    public static final class UnknownOperation extends Operation<OperationRequest> {
```
smiklosovic: i have to delete this

PR #201, opened 2019-07-24
Remove slash from leader ConfigMap name

johananl:
The slash causes startup errors. This looks like a typo to me.


alourie:
Sure, I think that was a mistake. I haven't tried to run the operator in the cluster though, so can't test whether it works or not without the fix, but we probably better off merging that.

:+1:  LGTM from me.

PR #202, opened 2019-07-24
Add a basic Dockerfile

johananl:
## General

This PR adds a basic Dockerfile for the new Go version, as well as a `.dockerignore` file to avoid including irrelevant files in the Docker build context, which tends to make Docker builds slower.

I've added the Dockerfile to the root of the repository since it's the one which builds the operator itself and that's a common practice in the OSS community. I'm aware of the [docker](https://github.com/instaclustr/cassandra-operator/tree/master/docker) directory which currently contains a dummy Dockerfile for the operator. I'd like to get feedback on whether we should move the Dockerfile to a different path.

## Implementation

This Dockerfile uses a multi-stage Docker build to make the final image as small as possible (using `scratch` as a base image). This not only makes the image smaller but also reduces the attack surface of the operator when running in a cluster. On the other hand this type of image can make debugging the operator harder since you don't have a shell and basic utilities, so please share your feedback on this.

## Teseting

To build the image, run `docker build -t cassandra-operator .`, then use the resulting image when deploying the operator to a cluster. Alternatively, run locally using `docker run -it --rm cassandra-operator ./cassandra-operator`.


benbromhead:
Whats the benefit of building the operator image this way vs leveraging the makefile (which builds the images using the dockerfiles in the docker directory)?

johananl:
@benbromhead I've added relevant info in https://github.com/instaclustr/cassandra-operator/issues/204.

Right now the Dockerfile at https://github.com/instaclustr/cassandra-operator/tree/superdupertopsecretrewrite/docker/cassandra-operator is a dummy one. The current directory structure makes it tricky to get the Go source into the image. You would have to either do Docker build context manipulations (which harm build reproducibility) or copy binaries around, which IMO isn't great.

Additionally (and regardless the path issues), this Dockerfile produces a very thin image which contains just the binary. This is improves download time and security.

alourie:
I like the way this builds the binary. We might want to look to build the sidecar the same way.

OTOH, we _do_ have 2 containers to build (and possibly 3 if counting Cassandra), so having just 1 file in the root location doesn't sound right. We can think of a way to put them into a docker folder hierarchy still.

johananl:
>OTOH, we do have 2 containers to build (and possibly 3 if counting Cassandra), so having just 1 file in the root location doesn't sound right. We can think of a way to put them into a docker folder hierarchy still.

Yes, I've been thinking about this. My rationale for putting the operator's Dockerfile in the root of the repository was that this is the "main" Dockerfile for the project whereas the others belong to *dependencies*.

Anyway, sure - we can definitely keep the current directory structure + Makefile hierarchy if this makes more sense to you guys. I'll push a change in a moment. If this looks better, I'll squash the commits.

johananl:
By the way, one downside of the thin multi-stage build is that Docker caching doesn't work out of the box, which results in slower builds compared to single-stage builds.

johananl:
I've added another commit which adds version info to the operator binary. This one relies on this PR so I included it here. I can also move this to a separate PR if it's better this way.

johananl:
Closed in favor of https://github.com/instaclustr/cassandra-operator/pull/206.

PR #205, opened 2019-07-26
Initial work for clouds secrets and user-defined configmap

alourie:
WIP, can be used to push backup secrets


examples/go/example-datacenter.yaml
```diff
@@ -8,8 +8,10 @@ spec:
   cluster: "test-cluster"
   nodes: 3
   cassandraImage: "gcr.io/cassandra-operator/cassandra:3.11.3"
-  sidecarImage: "gcr.io/cassandra-operator/cassandra-sidecar:latest"
+  sidecarImage: "192.168.56.104:5000/cassandra-sidecar:latest"
```
zegelin: Revert.

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -101,13 +100,16 @@ func newPodSpec(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, volumes []co
 }
 
 func newCassandraContainer(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, dataVolumeClaim *corev1.PersistentVolumeClaim, configVolume *corev1.Volume) *corev1.Container {
-	const OperatorConfigVolumeMountPath = "/tmp/operator-config"
+	const (
+		OperatorConfigVolumeMountPath = "/tmp/operator-config"
+		UserConfigVolumePath = "/etc/cassandra/cassandra.yaml.d/003-user-overrides.yaml"
```
zegelin: This should be `/tmp/user-config` (and renamed to `...MountPath`), since the user ConfigMap and VolumeSource can contain multiple files.

The entry point script will then take care of moving these all to underneath `/etc/cassandra`

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
```diff
@@ -10,18 +10,17 @@ import (
...
-
-	Resources v1.ResourceRequirements `json:"resources"`
-
+	Cluster             string                       `json:"cluster,omitempty"`
+	Nodes               int32                        `json:"nodes"`
+	CassandraImage      string                       `json:"cassandraImage"`
+	SidecarImage        string                       `json:"sidecarImage"`
+	ImagePullPolicy     v1.PullPolicy                `json:"imagePullPolicy"`
+	ImagePullSecrets    []v1.LocalObjectReference    `json:"imagePullSecrets,omitempty"`
+	BackupSecrets       string                       `json:"backupSecrets,omitempty"`
```
zegelin: `ImagePullSecrets` are `v1.LocalObjectReference`s. Should this be the same?

alourie: yes, changing

pkg/controller/cassandradatacenter/statefulset.go
alourie: :+1:

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
alourie: redesigned to use secret volumes

alourie:
@z Latest allows setting env vars for various purposes including cloud providers creds. Also, restored `PrivilegedSupported`. Also now works with custom userConfigMaps.

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -165,7 +212,7 @@ func newSysctlLimitsContainer(cdc *cassandraoperatorv1alpha1.CassandraDataCenter
 		Image:           cdc.Spec.CassandraImage,
 		ImagePullPolicy: cdc.Spec.ImagePullPolicy,
 		SecurityContext: &corev1.SecurityContext{
-			Privileged: func() *bool { b := true; return &b }(),
+			Privileged: boolRef(cdc.Spec.PrivilegedSupported),
```
johananl: Is there a point in running the sysctl init container unprivileged? Shouldn't we completely drop the container if `cdc.Spec.PrivilegedSupported == false`?

alourie: That's a very good question. In Java version we did [this](https://github.com/instaclustr/cassandra-operator/blob/master/java/operator/src/main/java/com/instaclustr/cassandra/operator/controller/DataCenterReconciliationController.java#L340), so I assumed we might want to do this here too. The cluster seems to be working fine in my local testing, but I'm not sure it actually tests this flow properly.

johananl: Here is what happens when the sysctl container runs unprivileged:

```
kc logs -f cassandra-test-dc-cassandra-0 -c sysctl-limits 
+ sysctl -w vm.max_map_count=1048575
sysctl: permission denied on key 'vm.max_map_count'
+ true
```

The pod starts fine because we're doing `|| true`, however it has no effect in this case as it cannot modify the setting on the host.

AFAICT we can drop this container when we can't run as privileged.

alourie:
Superseded by #218

PR #207, opened 2019-07-26
Fix Docker image name handling

johananl:
The current state generates tags such as `registry/repocassandra` instead of `registry/repo/cassandra`.
I'm assuming there should be slashes here from looking at the other Makefiles, for example [this one](https://github.com/instaclustr/cassandra-operator/blob/cfb76a760d0b6f1c5dcc272c758adb18b5109254/docker/cassandra-sidecar/Makefile#L19-L20).

### DIFF
--- a/buildenv/build
+++ b/buildenv/build
@@ -9,6 +9,6 @@ cd `dirname "$BASH_SOURCE"`/..
 
 source ./buildenv/repo
 
-echo "Building ${REPO}:${TAG}"
+echo "Building ${REGISTRY}:${TAG}"
 
-docker build --tag "${REPO}:${TAG}" --tag "${REPO}:latest" --tag "$1" docker/${1} 1>/dev/null
+docker build --tag "${REGISTRY}:${TAG}" --tag "${REGISTRY}:latest" --tag "$1" docker/${1} 1>/dev/null
--- a/buildenv/build-all
+++ b/buildenv/build-all
@@ -3,7 +3,7 @@
 
 cd `dirname "$BASH_SOURCE"`/..
 
-export DOCKER_REPO="${REPO:-gcr.io/cassandra-operator}"
+export DOCKER_REGISTRY="${REGISTRY:-gcr.io/cassandra-operator/}"
 
 cd docker
 make
--- a/buildenv/build-cassandra
+++ b/buildenv/build-cassandra
@@ -11,6 +11,6 @@ source ./buildenv/repo
 
 echo "Building Cassandra versions in version list"
 while IFS="" read -r line || [[ -n "$line" ]]; do
-    echo "Building ${REPO}:${line}"
-    docker build --tag "${REPO}:${line}" --tag "${1}:${line}" --build-arg cassandra_version="${line}" docker/${1} 1>/dev/null
+    echo "Building ${REGISTRY}:${line}"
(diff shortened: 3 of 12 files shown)


alourie:
@z I agree we should add a comment, but also make them the same. For example, the base-jre image includes the slashes. If I define the repo with a slash, building the base-jre fails, and if I define the DOCKER_REPO without the slash at the end, the cassandra build then fails:

```bash
...
[INFO] ------------------------------------------------------------------------                                                                                               
[INFO] BUILD SUCCESS                                                                                                                                                          
[INFO] ------------------------------------------------------------------------                                                                                               
[INFO] Total time: 0.955 s                                                                                                                                                    
[INFO] Finished at: 2019-07-29T13:46:35+09:30                                                                                                                                 
[INFO] ------------------------------------------------------------------------                                                                                               
docker build \                                                                                                                                                                
        --build-arg cassandra_version=3.11.3 \                                                                                                                                
        --build-arg cassandra_k8s_addons_jar=cassandra-k8s-addons-1.0-SNAPSHOT.jar \                                                                                          
        --build-arg openjre_image="192.168.56.104:5000/base-openjre:stretch-"8u212-b01-1-deb9u1"" \                                                                           
        -t 192.168.56.104:5000cassandra \                                                                                                                                     
        -t 192.168.56.104:5000cassandra:3.11.3 \                                                                                                                              
        .                                                                                                                                                                     
invalid argument "192.168.56.104:5000cassandra:3.11.3" for "-t, --tag" flag: invalid reference format                                                                         
See 'docker build --help'.                                                                                                                                                    
make[1]: *** [Makefile:18: cassandra] Error 125                                                                                                                               
make[1]: Leaving directory '/home/alourie/Projects/gospace/src/github.com/instaclustr/cassandra-operator/docker/cassandra'                                                    
make: *** [Makefile:8: cassandra] Error 2                                                                          
```

johananl:
@alourie wrong mention?

alourie:
@johananl nope, right mention ;-). We need to unify this stuff around different makefiles.

johananl:
@alourie I thought you meant to tag @zegelin (`@z` is a user named Tyler Mulligan who doesn't seem to be active in this repo, unless I'm missing something).

In any case, how should we proceed with this?

alourie:
@johananl yep, you're right. Fixed mention. I want to wait for Adam to have a take on this.

zegelin:
There should be no slash in any of the Makefiles. Meaning that if `$(DOCKER_REPO)` the resulting tag name will be `tag-name` (which builds), not `/tag-name` (which does not build).

I looked into this further, since the master branch doesn't have this problem.

We've actually diverged a bit it appears. Master supports LDAP and Kerberos too, and a bunch of other changes were made to fix CircleCI builds and stuff. I never brought the changes over to the rewrite branch.

johananl:
Following @zegelin's feedback, I've amended the commit to remove all slashes from the image names. I've also proposed changing `REPO` and `DOCKER_REPO` to `REGISTRY` and `DOCKER_REGISTRY`, respectively. Rationale is in the commit message.

We may want to document the correct usage of `DOCKER_REGISTRY` somewhere.

smiklosovic:
@johananl yes, please document that and we can merge this.

smiklosovic:
@johananl can you please rebase this on top of the current superduper? Thank you!

johananl:
> lgtm. Future additions could maybe be an error if `REGISTRY` is set but doesn't end in a slash...

Tracked in https://github.com/instaclustr/cassandra-operator/issues/226.

PR #215, opened 2019-07-29
Run k8s code generators

johananl:
Generated code isn't aligned with API spec, most likely because `operator-sdk generate k8s` wasn't run after implementing `CassandraBackup`.


pkg/apis/cassandraoperator/v1alpha1/zz_generated.openapi.go
```diff
@@ -1,6 +1,4 @@
-// +build !ignore_autogenerated
-
-// Code generated by openapi-gen. DO NOT EDIT.
+// +build !
```
johananl: Not sure why the generator made this change. The lonely `!` doesn't make sense to me. This could be a bug in https://github.com/kubernetes/gengo.

Any ideas?

johananl: Bug reported in https://github.com/operator-framework/operator-sdk/issues/1747.

alourie:
@johananl  I thought I ran all of it, apparently not. Do we need to do anything else on top of you running it?

johananl:
>Do we need to do anything else on top of you running it?

I would consider waiting at least until we get a response on https://github.com/operator-framework/operator-sdk/issues/1747 because of https://github.com/instaclustr/cassandra-operator/pull/215#discussion_r308383610. The code builds fine in the current state but to me the single `!` in the build tag definitely looks like a bug in the operator framework, which could potentially create unexpected behavior.

I'm not sure if/how the above actually influences the operator. I could spend some more time on it this week to figure out if it's actually a problem (or even fix the bug).

johananl:
https://github.com/operator-framework/operator-sdk/pull/1757 is pending merge. I was able to confirm this PR fixes the bug. The PR is currently blocked by a flaky CI process, and in any case it will likely take a while until a new version of operator-sdk is released.

Should we fix the build tag manually as a workaround and merge it? I would like to advance https://github.com/instaclustr/cassandra-operator/pull/216 which is blocked by this one.

johananl:
I've pushed a workaround commit. Please feel free to discard it if you guys would like to wait for the fix instead.

We can probably get by by using `opeartor-sdk` from https://github.com/operator-framework/operator-sdk/pull/1757 for code generation until the fix is released.

johananl:
https://github.com/operator-framework/operator-sdk/pull/1757 is merged (but not released yet).

smiklosovic:
@johananl any updates here?

smiklosovic:
I see it is already merged into superduper? Do I see it right? I am wondering how that endup there ...

johananl:
Generators were already run in https://github.com/instaclustr/cassandra-operator/pull/216/commits/e392126cc4cb485852830c458a70fe69638e6f1a so we are now aligned in the base branch. Closing.

PR #216, opened 2019-07-29
Add serviceAccountName field to CassandraDataCenter

johananl:
## General

This PR adds support for specifying a k8s ServiceAccount to be used for pods which the operator creates.

This is important in cases where the operator is deployed on clusters with PodSecurityPolicies which don't allow the C* pods to run, in which case the cluster admin should have the flexibility to allow using an existing PSP on the cluster which allows C* to run.

>NOTE: This PR requires re-running `operator-sdk generate`. However, the current state in the base branch isn't aligned. This causes irrelevant changes to `zz_generated.*` to be included in this PR. We should only merge this PR once https://github.com/instaclustr/cassandra-operator/pull/215 is merged (and rebase before we do).

## Testing

To test this PR, compile and run the operator on a k8s cluster, then create the following resources:

```yaml
# A sample PSP which allows Cassandra pods to run
apiVersion: policy/v1beta1
kind: PodSecurityPolicy
metadata:
  name: cassandra
  annotations:
    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default'
    seccomp.security.alpha.kubernetes.io/defaultProfileName:  'docker/default'
spec:
  privileged: true
  allowPrivilegeEscalation: true
  allowedCapabilities:
  - IPC_LOCK
  - SYS_RESOURCE
  requiredDropCapabilities:
  - KILL
  - MKNOD
  - SETUID
  - SETGID
  volumes:
  - 'configMap'
  - 'emptyDir'
  - 'projected'
  - 'secret'
  - 'downwardAPI'
  - 'persistentVolumeClaim'
  hostNetwork: false
  hostIPC: false
  hostPID: false
  runAsUser:
    rule: 'RunAsAny'
  seLinux:
    rule: 'RunAsAny'
  supplementalGroups:
    rule: 'MustRunAs'
    ranges:
    - min: 1
      max: 65535
  fsGroup:
    rule: 'MustRunAs'
    ranges:
    - min: 1
      max: 65535
  readOnlyRootFilesystem: false
```

```yaml
# A role which is allowed to use the PSP above
apiVersion: rbac.authorization.k8s.io/v1
kind: Role
metadata:
  creationTimestamp: null
  name: cassandra
rules:
- apiGroups: ['policy']
  resources: ['podsecuritypolicies']
  verbs:     ['use']
  resourceNames:
  - cassandra
```

```yaml
# A ServiceAccount which is allowed to use the role above
apiVersion: v1
kind: ServiceAccount
metadata:
  name: cassandra
```

```yaml
# A RoleBinding which allows the cassandra SA to use the cassandra role
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  name: cassandra
subjects:
  - kind: ServiceAccount
    name: cassandra
roleRef:
  kind: Role
  name: cassandra
  apiGroup: rbac.authorization.k8s.io
```

```yaml
# A Cassandra deployment with the new field
apiVersion: cassandraoperator.instaclustr.com/v1alpha1
kind: CassandraDataCenter
metadata:
  name: test-dc-cassandra
  labels:
    app: cassandra
spec:
  cluster: "test-cluster"
  nodes: 3
  # Modify Docker image tags to the images you've built
  cassandraImage: "<repo>/cassandra"
  sidecarImage: "<repo>/cassandra-sidecar"
  imagePullPolicy: Always
  resources:
    limits:
      memory: 1Gi
    requests:
      memory: 1Gi

  dataVolumeClaimSpec:
    accessModes:
      - ReadWriteOnce
    resources:
      requests:
        storage: 500Mi

  prometheusSupport: false
  privilegedSupported: true

  serviceAccountName: cassandra
```

Run the following command to check the SA on the created pods:

```
kubectl get pods cassandra-test-dc-cassandra-0 -o=jsonpath='{.spec.serviceAccountName}'
```

The output should be `cassandra`.

## TODO

- [ ] Add documentation
- [ ] Add tests


smiklosovic:
@johananl good stuff! I like this ... seems to me like the way we should go with.

Anyway, can you check what user / group we run Cassandra under? I think it is root right now or am I mistaken? Maybe we should further restrict this to `cassandra:cassandra`? Is that in principle possible?

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -91,10 +92,11 @@ func newStatefulSetSpec(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, podS
...
 	podSpec := &corev1.PodSpec{
-		Volumes:          volumes,
-		Containers:       containers,
-		InitContainers:   initContainers,
-		ImagePullSecrets: cdc.Spec.ImagePullSecrets,
+		Volumes:            volumes,
+		Containers:         containers,
+		InitContainers:     initContainers,
+		ImagePullSecrets:   cdc.Spec.ImagePullSecrets,
+		ServiceAccountName: cdc.Spec.ServiceAccountName,
```
alourie: What happens if ServiceAccountName is not set in the CRD? Will this work with empty string?

johananl:
>Anyway, can you check what user / group we run Cassandra under? I think it is root right now or am I mistaken? Maybe we should further restrict this to cassandra:cassandra? Is that in principle possible?

@smiklosovic indeed, both Cassandra and the sidecar currently run as root. I've opened https://github.com/instaclustr/cassandra-operator/pull/219 to make them run as non-root.

pkg/controller/cassandradatacenter/statefulset.go
johananl: Yes, we are using `omitempty` in the spec so omitting this field simply creates the StatefulSet without the `ServiceAccountName` field. It's the same as the existing `ImagePullSecrets` field.

https://github.com/instaclustr/cassandra-operator/pull/216/files#diff-ef654dce75ffebdf8a5ce52918b76039R27

PR #217, opened 2019-07-30
backup-restores wip

smiklosovic:
I want to get rid of this so we can iterate on that. Notes and stuff are in the project discussions ... Dont obsess and do any bikeshedding here please. I will squash afterwards.


java/backup-restore/pom.xml
```diff
@@ -0,0 +1,153 @@
...
+        </dependency>
+
+        <dependency>
+            <groupId>com.google.inject.extensions</groupId>
+            <artifactId>guice-assistedinject</artifactId>
+        </dependency>
+
+        <dependency>
+            <groupId>com.datastax.cassandra</groupId>
+            <artifactId>cassandra-driver-core</artifactId>
```
benbromhead: can we make sure we are explicitly versioning our dependencies, otherwise the build may randomly break when the latest version is upgraded

java/backup-restore/src/main/java/com/instaclustr/cassandra/backup/azure/AzureBackuper.java
```diff
@@ -1,41 +1,47 @@
...
-    public AzureSnapshotUploader(final CloudBlobClient cloudBlobClient,
-                                 final BackupArguments arguments) throws URISyntaxException, StorageException {
-        super(arguments.clusterId, arguments.backupId, arguments.backupBucket);
+    @Inject
+    public AzureBackuper(final CloudBlobClientProvider cloudBlobClientProvider,
+                         final Executors.ExecutorServiceSupplier executorServiceSupplier,
+                         @Assisted final BackupOperationRequest request) throws Exception {
+        super(request, executorServiceSupplier);
 
         //Currently just use clusterId (name) as container reference
```
benbromhead: Based on your changes here I don't think this old comment still applies?

java/backup-restore/src/main/java/com/instaclustr/cassandra/backup/cli/CommitLogBackupApplication.java
```diff
@@ -0,0 +1,41 @@
...
+import org.slf4j.LoggerFactory;
+import picocli.CommandLine;
+
+@CommandLine.Command(name = "commitlog-backup",
+        mixinStandardHelpOptions = true,
+        description = "Upload archived commit logs to remote storage.",
+        sortOptions = false,
+        versionProvider = CLIJarManifestVersionProvider.class
+)
+public class CommitLogBackupApplication implements Runnable {
```
benbromhead: Does this now support restoring commitlog segments?

java/k8s-addons/src/main/java/com/instaclustr/cassandra/k8s/SeedProvider.java
```diff
@@ -1,14 +1,14 @@
 package com.instaclustr.cassandra.k8s;
 
-import com.google.common.collect.ImmutableList;
```
benbromhead: This comment is non-blocking on the PR and doesn't need to get fixed, but for the future, try not to check in unrelated changes to the feature branch. Especially if the IDE is just redoing imports :)

java/backup-restore/pom.xml
smiklosovic: yes, it is versioned, that dep (with version) is just in `dependencyManagement` in parent so there is not a need to have it with versions here

java/backup-restore/src/main/java/com/instaclustr/cassandra/backup/cli/CommitLogBackupApplication.java
smiklosovic: I am not so far with this but yes, that is the goal ...

java/backup-restore/pom.xml
benbromhead: Awesome!!!

PR #218, opened 2019-08-02
Support user configs, user secrets and separate environments for cassandra and sidecar

### LINKED ISSUE #178: Port secret/keystore support
Currently the Java based operator allows users to specify secrets to embed in the C* pods and provide custom configuration for the nodes to use those keystores. This allows users to setup TLS encryption for internode encryption and client-node encryption. 

We need to ensure the operator-sdk based version also supports this workflow

### LINKED ISSUE #213: Port user config map support.
The Java project supports user-specified ConfigMaps via the `userConfigMapVolumeSource` field of the `CassandraDataCenter` CRD.

alourie:
Fixes #213


examples/go/example-datacenter.yaml
```diff
@@ -10,6 +10,31 @@ spec:
...
+    items:
+      - key: creds.json
+        path: creds.json
+  env:
+    - name: GOOGLE_APPLICATION_CREDENTIALS
+      value: "/etc/google/creds.json"
+  userSecretVolume:
+    secretName: dc1-user-secret
+  userConfigMapVolumeSource:
+    name: dc1-user-config
```
zegelin: Where is this config-map defined? The example should be complete.

examples/go/example-datacenter.yaml
```diff
@@ -10,6 +10,31 @@ spec:
...
+  backupSecretVolume:
+    secretName: backup-secret
+    #     type is a workaround for https://github.com/kubernetes/kubernetes/issues/68466
+    type: array
+    items:
+      - key: creds.json
+        path: creds.json
+  env:
+    - name: GOOGLE_APPLICATION_CREDENTIALS
+      value: "/etc/google/creds.json"
```
zegelin: Is this the "default"? If so, do we need to set it as an environment variable?

examples/go/example-datacenter.yaml
```diff
@@ -10,6 +10,31 @@ spec:
...
+    name: dc1-user-config
+    #     type is a workaround for https://github.com/kubernetes/kubernetes/issues/68466
+    type: array
+    items:
+      - key: cassandra_ssl
+        path: cassandra.yaml.d/003-ssl.yaml
+      - key: cqlshrc
+        path: cqlshrc
+      - key: install_cqlshrc
+        path: cassandra-env.sh.d/003-install-cqlshrc.sh
```
zegelin: This seems hacky... `cassandra-env.sh` is for setting up the JVM and its environment. Not other tools.

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -157,6 +191,30 @@ func newSidecarContainer(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, dat
 			{Name: podInfoVolume.Name, MountPath: "/etc/pod-info"},
 		},
 	}
+
+	if len(cdc.Spec.Env) > 0 {
+		container.Env = cdc.Spec.Env
+	}
+
+	if backupSecretVolume != nil {
```
zegelin: I think it'd be preferable to not have the operator deal with provider-specific stuff unless 100% necessary.
This code also seems to imply that the backup destination would always be GCP.

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -165,7 +223,7 @@ func newSysctlLimitsContainer(cdc *cassandraoperatorv1alpha1.CassandraDataCenter
 		Image:           cdc.Spec.CassandraImage,
 		ImagePullPolicy: cdc.Spec.ImagePullPolicy,
 		SecurityContext: &corev1.SecurityContext{
-			Privileged: func() *bool { b := true; return &b }(),
+			Privileged: boolRef(cdc.Spec.PrivilegedSupported),
```
zegelin: Can't this just be `&cdc.Spec.PrivilegedSupported`? The old lambda was to work around that you can'd to `&true`.

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
```diff
@@ -10,18 +10,20 @@ import (
...
-	DataVolumeClaimSpec v1.PersistentVolumeClaimSpec `json:"dataVolumeClaimSpec"`
-
-	PrometheusSupport bool `json:"prometheusSupport"`
+	Cluster                   string                       `json:"cluster,omitempty"`
+	Nodes                     int32                        `json:"nodes"`
+	CassandraImage            string                       `json:"cassandraImage"`
+	SidecarImage              string                       `json:"sidecarImage"`
+	ImagePullPolicy           v1.PullPolicy                `json:"imagePullPolicy"`
+	ImagePullSecrets          []v1.LocalObjectReference    `json:"imagePullSecrets,omitempty"`
+	BackupSecretVolume        v1.SecretVolumeSource        `json:"backupSecretVolume,omitempty"`
```
zegelin: `BackupSecretVolume` and `UserSecretVolume` should be named `...VolumeSource` for consistency with `UserConfigMapVolumeSource`.

If these fields are optional, they should be pointers.

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -100,16 +126,15 @@ func newPodSpec(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, volumes []co
...
-
+func newCassandraContainer(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, dataVolumeClaim *corev1.PersistentVolumeClaim, configVolume *corev1.Volume, userSecretVolume, userConfigVolume *corev1.Volume) *corev1.Container {
 	container := &corev1.Container{
 		Name:            "cassandra",
 		Image:           cdc.Spec.CassandraImage,
 		ImagePullPolicy: cdc.Spec.ImagePullPolicy,
 		Args:            []string{OperatorConfigVolumeMountPath},
 		Ports: []corev1.ContainerPort{
 			{Name: "internode", ContainerPort: 7000},
+			{Name: "internode-ssl", ContainerPort: 7001},
```
zegelin: `tls` ;)

examples/go/example-datacenter.yaml
smiklosovic: There does not seem to be any default https://cloud.google.com/docs/authentication/getting-started

smiklosovic: i checked the source code, there is not any default location

alourie: yep, no defaults for that one. This is just an example of where to place that file, I literally look it up in the env and if it's there - we use it. If not defined, we'll use `/etc/gcp`, but I might change that to something more generic.

pkg/controller/cassandradatacenter/statefulset.go
alourie: :man_facepalming:

pkg/apis/cassandraoperator/v1alpha1/cassandradatacenter.go
alourie: yes, I've thought about it after pushing, and was already too sleepy to think about it too much, so sure, will rename.

pkg/controller/cassandradatacenter/statefulset.go
alourie: :+1:

examples/go/example-datacenter.yaml
alourie: I basically copied the example from the original java code. I'll probably copy it back to the `doc` folder to add to one of the examples.

examples/go/example-datacenter.yaml
alourie: sure, I'll think of a better usecase.

pkg/controller/cassandradatacenter/statefulset.go
alourie: As per above, I'll think a more generic location. Clearly that this was done initially for GCP, and other providers will probably not use it....but of course they might.

examples/go/example-datacenter.yaml
alourie: added doc

examples/go/example-datacenter.yaml
alourie: removed, decided to use an environment variable and move it to the ~/.cassandra location with entry-point.

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -24,21 +23,9 @@ const (
 	OperatorConfigVolumeMountPath = "/tmp/operator-config"
 	UserConfigVolumeMountPath     = "/tmp/user-config"
 	UserSecretVolumeMountPath     = "/tmp/user-secret-config"
+	BackupSecretVolumeMountPath   = "/tmp/backup-secret"
```
zegelin: What's the difference between `BackupSecret` and `UserSecret`?

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -24,21 +23,9 @@ const (
 	OperatorConfigVolumeMountPath = "/tmp/operator-config"
 	UserConfigVolumeMountPath     = "/tmp/user-config"
 	UserSecretVolumeMountPath     = "/tmp/user-secret-config"
```
zegelin: This probably should just be `/tmp/user-secret` for consistency.

docker/cassandra/entry-point
```diff
@@ -12,4 +12,12 @@ do
 done
 )
 
+# Copy over cqlshrc to a default location if defined as ENV
```
zegelin: I think we can make this nicer. It IMO shouldn't be a requirement that users have to define an environment variable and add a file to a config map to get cqlsh to function.

Instead I suggest that we write our own wrapper for `cqlsh` that detects if SSL is enabled, and fetches the CA cert (or just disables trust checking). This will also be necessary once we support authn/authz, since we need to implement shared-secret support.

As a side question, should `cqlsh` be installed the C* container at all? If people want to connect to the cluster, should they instead start a new pod/container that just contains cqlsh?

examples/go/example-datacenter.yaml
zegelin: Looks like the default process is this: https://cloud.google.com/docs/authentication/production#finding_credentials_automatically

So yes, we need a way to define the environment variable, but that's probably also going to be the more uncommon approach, as on GKE the credentials will be automatically discovered.

C* clusters running on GKE would most likely backup to Google Cloud Storage. Same on Amazon EKS -- S3 is the most likely destination. But we don't want to prevent clusters running on other K8s environments from accessing cloud storage, so having the option is always nice. Could also be useful for GKE -> EKS migrations for example.

pkg/controller/cassandradatacenter/statefulset.go
alourie: The difference is in fact that `BackupSecret` would hold credentials for backup functionality (i.e. GCP) and be mounted in sidecar container, while `UserSecret` is for C* encryption/auth credentials and to be mounted in the Cassandra container. Having backup stuff mounted in the cassandra container or cassandra credentials in the sidecar container, while not horrible, feels pretty hacky and dirty. That's why the split.

docker/cassandra/entry-point
alourie: @zegelin we currently use cqlsh now in `cql-readiness-probe`, and this is easily replaceable with either python script or, say, a statically compiled gocql simple program. So we don't have an inherent requirement to have it in the container at all.

Also, if people want to use cqlsh to connect to the cluster, they can either spin a new container with all the certificates and custom configs as they need to, or have it installed locally if they have the network access to the cluster. So, if you think that's a good path to follow, I can replace the readiness script with something else pretty quickly and stop the dependency on `cqlsh`.

examples/go/example-datacenter.yaml
alourie: Yea, it's a convenient option, and it's optional. Also, might be that even on GKE they might want to use some different account for backups, and that's when it would be handy.

docker/cassandra/entry-point
alourie: I did rush it a bit. Clearly, we need any "script" to know about SSL configuration to be able to connect to the node. So having a wrapper around cqlsh might be a useful and simpler solution.

pkg/controller/cassandradatacenter/statefulset.go
alourie: :+1:

zegelin:
Is this ready to be merged?

alourie:
@zegelin it can be merged, but the "secure" Cassandra needs more work. I'd prefer merging this, as it fixes a bunch of tickets/features/bugs, and continuing work on proper TLS support separately. If we do it, I'll rename the merge to "Support backups secrets, user secrets and user maps" to avoid misunderstanding.

pkg/controller/cassandradatacenter/statefulset.go
benbromhead: Maybe change to sidecarSecret/cassandraSecret? So it relates to the pod name

I don't mind the C* and sidecar containers sharing secrets to be honest.

docker/cassandra/entry-point
benbromhead: In order to get the PR closed, I think we should create a new issue/PR for the wrapper script and solve that one separately.

alourie: yea, I'm going to remove this part and deal with that in #241

pkg/controller/cassandradatacenter/statefulset.go
alourie: Well, I can rename it, but I think it's better called a "backup" secret as it hints the usage for backups only.

Actually, the only other "secrets" usage would be for secure communications as far as I can tell, and we're going to resolve it differently anyway. So I'd just keep it as "backup" for now, and if ever we combine "user" and "backup" secrets, we can change the naming convention.

smiklosovic:
looks good to me

alourie:
Now can be squashed and merged @benbromhead

Issue #208, opened 2019-07-26
Setting sysctl parameters on nodes


johananl:
This issue is partially discussed in https://github.com/instaclustr/cassandra-operator/issues/116.

## General

At the moment the Go version of the operator sets `vm.max_map_count` on worker nodes by running a privileged init container:

https://github.com/instaclustr/cassandra-operator/blob/cfb76a760d0b6f1c5dcc272c758adb18b5109254/pkg/controller/cassandradatacenter/statefulset.go#L162-L175

Doing this by default can be unexpected for a user as this is a host-level setting which could potentially affect other pods on the relevant nodes. In addition, some environments may not allow running privileged containers, which can make deployments break.

The examples seem to [hint](https://github.com/instaclustr/cassandra-operator/blob/superdupertopsecretrewrite/examples/go/example-datacenter.yaml#L27) this is an optional feature, however this field isn't included in the `CassandraDataCenter` API, which means it is ignored by the operator.

## Suggested Changes

Assuming this setting is indeed necessary, I would consider the following:

- Make this setting **optional** and **off by default**.
- Mention in the docs that Cassandra works best when setting this option, and let the user decide whether they want to ignore it, use the provided init container or set the option at the infrastructure level.
- Rename the option from `privilegedSupported` to something which better describes what it does. Example: `optimizeSysctl`.

Implementing the suggestions above would make less assumptions on the user's environment and avoid taking decisions on behalf of the user which may exceed the responsibility of this operator.

zegelin:
`privilegedSupported` being missing on the Go side is indeed a regression. @alourie we need to add this.

Also agree that perhaps `privilegedSupported` isn't the best name, and that something like `optimizeSysctl` or `tuneSysctls` would be better.

The original name is derived from that fact that to call `sysctl` the security context `Privileged` flag needs to be set to true.

zegelin:
> Agree with it being off by default to avoid surprises.

Actually, I'd like to open this for discussion with @benbromhead @smiklosovic 

Cassandra really likes to create a large number of memory mappings, and doesn't handle things gracefully when this fails.

I can see a few options:

- Default to `true`. This results in Pod creation failure on clusters that don't support privileged containers. This gives upfront discovery of the problem, rather than having C* be unstable down the road. When supported, you get (potentially) higher performance clusters. The downside is that things may not work out-of-the-box, and we may end up with frequent issues being raised (or "me too"'ed) around this problem.

[...]

PR #219, opened 2019-08-02
Run Cassandra and sidecar as non-root

johananl:
This PR makes C* and the sidecar run as a non-root user. Rationale:

- Running containers as root is less secure than non-root.
- Some clusters employ PSPs which prevent running containers as root. This change makes it possible to use the operator even on clusters with strict PSPs.
- As far as I can tell, root isn't really necessary for C* nor for the sidecar.

### DIFF
--- a/docker/cassandra/Dockerfile
+++ b/docker/cassandra/Dockerfile
@@ -19,6 +19,7 @@ COPY cql-readiness-probe /usr/bin/cql-readiness-probe
 COPY ${cassandra_k8s_addons_jar} /usr/share/cassandra/lib
 
 ADD default-config /etc/cassandra
+RUN chown -R cassandra:cassandra /etc/cassandra
 
 VOLUME /var/lib/cassandra
 VOLUME /etc/cassandra
@@ -29,5 +30,11 @@ VOLUME /etc/cassandra
 # 9042: CQL
 # 9160: thrift service
 EXPOSE 7000 7001 7199 9042 9160
+
+# Run as user 'cassandra'
+# A numeric UID is used for PSP support:
+# https://kubernetes.io/docs/concepts/policy/pod-security-policy/#users-and-groups
+# The user is created in ./install-cassandra
+USER 999
 
 ENTRYPOINT ["/usr/bin/entry-point"]
--- a/docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
+++ b/docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
@@ -1,1 +1,0 @@
-ulimit -l unlimited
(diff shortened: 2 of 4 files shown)


docker/cassandra/Dockerfile
```diff
@@ -7,6 +7,10 @@ ARG cassandra_k8s_addons_jar
 
 COPY install-cassandra /tmp/install-cassandra
 
+# Create 'cassandra' user and group
```
zegelin: Lets move the user & group add and the chown back into `install-cassandra.

docker/cassandra/Dockerfile
```diff
@@ -19,6 +23,7 @@ COPY cql-readiness-probe /usr/bin/cql-readiness-probe
 COPY ${cassandra_k8s_addons_jar} /usr/share/cassandra/lib
 
 ADD default-config /etc/cassandra
+RUN chown -R cassandra:cassandra /etc/cassandra
 
 VOLUME /var/lib/cassandra
```
zegelin: What about permissions on the `/var/lib/cassandra` volume? Is the default owner root?

johananl: No, they are owned by `cassandra` already (i.e. even before this change):

```
root@0ff4dd9f253d:/# ls -alh /var/lib/ | grep cas
drwxr-xr-x  2 cassandra cassandra 4.0K Jul  2  2018 cassandra
```

docker/cassandra/Dockerfile
johananl: Done.

docker/cassandra/default-config/cassandra-env.sh.d/002-cassandra-limits.sh
```diff
@@ -1 +0,0 @@
-ulimit -l unlimited
```
johananl: Running `ulimit` doesn't seem to be possible in k8s: https://github.com/kubernetes/kubernetes/issues/3595

Do we really need to run `ulimit -l unlimited` inside the container? AFAIK this is only relevant when swap is enabled, which doesn't seem to be very common in k8s clusters.

In case this *is* necessary for a production-ready C* deployment, following [this](https://github.com/instaclustr/cassandra-operator/issues/208#issuecomment-516029789) approach I would consider omitting the command from the container image and instead recommend a host-level tweak in the docs, for example disabling swap on all worker nodes if C* is intended to serve production traffic.

zegelin: @johananl you’re probably right that `ulimit -l unlimited` isn’t necessary.

Some background: C* wants to call `mlockall` on startup to prevent the JVM heap (which gets sized so the min = max — it never grows) from ever being paged out. This is done to prevent GC pause times from sky-rocketing, as any GC that does a full heap pass would cause everything to be paged back in.

That is really only a concern AFAIK if swap is enabled.

If C* can’t call `mlockall`, it will just print a warning and move on.

alourie:
lgtm, I'm also in for removing `ulimit` command

Issue #227, opened 2019-08-09
Update operator-sdk to v0.10.0


johananl:
operator-sdk has [v0.10.0](https://github.com/operator-framework/operator-sdk/releases/tag/v0.10.0) already. We are currently at v0.7.0. We should upgrade.

johananl:
Done in https://github.com/instaclustr/cassandra-operator/pull/229.

PR #229, opened 2019-08-09
Update operator-sdk to v0.10.0

johananl:
Fixes https://github.com/instaclustr/cassandra-operator/issues/227.

This PR updates to operator-sdk `v0.10.0`. It also moves from dep to Go modules.

>~NOTE: I am still testing this, but opening for review already since it's a big one.~

## Testing

```
cd /tmp
git clone https://github.com/kinvolk/cassandra-operator.git
cd cassandra-operator
git checkout operator-sdk-v0.10.0
cd cmd/manager
go build
```

When building the code inside `$GOPATH`, setting `export GO111MODULE=on` may be necessary. More info: https://github.com/golang/go/wiki/Modules


johananl:
Basic functionality works fine on my testing.

go.mod
```diff
@@ -0,0 +1,39 @@
...
+	gotest.tools v2.2.0+incompatible
+	k8s.io/api v0.0.0-20190612125737-db0771252981
+	k8s.io/apimachinery v0.0.0-20190612125636-6a5db36e93ad
+	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/kube-openapi v0.0.0-20190603182131-db7b694dc208
+	sigs.k8s.io/controller-runtime v0.1.12
+	sigs.k8s.io/controller-tools v0.1.10
+)
+
+// Pinned to kubernetes-1.13.4
```
smiklosovic: @johananl why this version? There is 1.15 already, is operator-sdk 0.10.0 known to work with this one?

smiklosovic:
@johananl thanks! faster than expected :) Lets keep this open for a while for @alourie to give a feedback too.

go.mod
johananl: Good question. This file is generated by the SDK. I'll have to check this.

pkg/controller/cassandrabackup/cassandrabackup_controller.go
```diff
@@ -47,23 +108,28 @@ func (r *ReconcileCassandraBackup) Reconcile(request reconcile.Request) (reconci
 	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
 	reqLogger.Info("Reconciling CassandraBackup")
 
-	// Get request params
-	b := &cassandraoperatorv1alpha1.CassandraBackup{}
-	if err := r.client.Get(context.TODO(), request.NamespacedName, b); err != nil {
+	// Fetch the CassandraBackup instance
+	instance := &cassandraoperatorv1alpha1.CassandraBackup{}
```
alourie: can we call this `backup` instead of `instance`? "Instance" is very ambiguous, and has a lot of meanings, which can be very confusing in a backup context.

johananl:
>1. We've "reconsolidated" reconciler/controller back to <>_controller; any specific benefit from doing so?
>2. We've added a status subresource. How does work together with a specific Status field in CRD (specifically in case of Backups)?

@alourie both of these changes come from upstream, i.e. this is how `operator-sdk` generates the code at its current version. I've merged these changes because I thought it makes sense to keep the "scaffolding" code up to date with the SDK. I didn't look too deep into the implications there and just trusted the SDK folks in this case. I can definitely look a little deeper into the implications of these changes, even if only to be able to justify them for ourselves.

pkg/controller/cassandrabackup/cassandrabackup_controller.go
johananl: `instance` comes from the SDK scaffolding code generator. True, it's generic, probably on purpose. I guess the SDK folks wanted `instance` to always mean "the resource which needs reconciling" if that makes sense.

We can change it to `backup` or back to `b`, however this would make it slightly harder to figure out the next time we do an SDK version upgrade. WDYT?

smiklosovic: I would leave it to be `instance` in this particular case.

alourie:
@johananl did you manually merge reconciler/controller or did operator-sdk knew to do it automatically?

In any case, lgtm, no need to look deeper into this at this stage.

johananl:
>@johananl did you manually merge reconciler/controller or did operator-sdk knew to do it automatically?

The SDK doesn't provide a way to automatically migrate code when upgrading to newer versions of the SDK. The recommendation is to generate the project from scratch using the new version and then copy the custom code into the new stubs.
Initially I thought we would get by with just updating the Go dependency and manually figuring out the code changes, but it quickly proved to be the wrong approach, especially since breaking changes were introduced between `v0.7.0` and `v0.10.0`. The upside is that we've switched to Go modules now :-)

[Here](https://github.com/operator-framework/operator-sdk/blob/51d28df214440d97031368d7d700a2805161015d/doc/migration/v0.1.0-migration-guide.md) is a migration guide for an older version. I've used a very similar process to get to the current version.

smiklosovic:
@johananl merged, thanks!

PR #239, opened 2019-08-14
Add a PSP for operator

johananl:
This PR adds a PodSecurityPolicy for the operator pod.

On some clusters some default PSP may prevent the operator from running. Shipping a PSP with the operator makes it easier for the cluster admin to figure out exactly what permissions are required by the operator, and provides a base for customizations.

## Testing

Deploy the operator to a k8s cluster as usual and apply `psp.yaml` using `kubectl` as well. Verify the PSP is indeed used by the operator by running the following:

```
kubect get pods --selector name=cassandra-operator -oyaml | grep psp
```

The output should be `kubernetes.io/psp: cassandra-operator`.

### DIFF
--- a/deploy/psp.yaml
+++ b/deploy/psp.yaml
@@ -1,0 +1,40 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: cassandra-operator
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: 'docker/default'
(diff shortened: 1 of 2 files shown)


smiklosovic:
@johananl thanks for this. I will try it all soonish.

smiklosovic:
pushed here https://github.com/instaclustr/cassandra-operator/commit/2372b2e048e4e2e657943ad4bda4eaf5096e17d3

PR #243, opened 2019-08-21
implementation of backup in go

smiklosovic:


java/backup-restore/src/main/java/com/instaclustr/cassandra/backup/aws/S3Backuper.java
```diff
@@ -169,7 +169,7 @@ public void cleanup() {
         try {
             // TODO cleanupMultipartUploads gets access denied, INS-2326 is meant to fix this
             cleanupMultipartUploads();
-
+            transferManager.shutdownNow(true);
```
zegelin: I think this call should be moved outside this try block, or into its own try block if exceptions should be handled.

Right now, if `cleanupMultipartUploads()` fails, which it can due to permissions problems, then the transfer manager won't be shut down cleanly.

smiklosovic:
merged here https://github.com/instaclustr/cassandra-operator/commit/f42757f854b04798edccb62e1280c555d3bff55f

PR #245, opened 2019-08-22
Improve Cassandra rack configuration handling

alourie:


pkg/controller/cassandradatacenter/configmap.go
```diff
@@ -60,14 +58,50 @@ func createOrUpdateOperatorConfigMap(rctx *reconciliationRequestContext, seedNod
 	return configVolume, nil
 }
 
-func addCassandraGossipingPropertyFileSnitchProperties(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, addFileFn func(path string, data string)) {
+func createOrUpdateCassandraRackConfig(rctx *reconciliationRequestContext, rack string) (*corev1.Volume, error) {
+
+	configMap := &corev1.ConfigMap{ObjectMeta: DataCenterResourceMetadata(rctx.cdc, "operator-rack-config")}
```
zegelin: Shouldn't the ConfigMap name contain the name of the rack? Otherwise this would overwrite when there is multiple racks, no?

pkg/controller/cassandradatacenter/configmap.go
```diff
@@ -60,14 +58,50 @@ func createOrUpdateOperatorConfigMap(rctx *reconciliationRequestContext, seedNod
 	return configVolume, nil
 }
 
-func addCassandraGossipingPropertyFileSnitchProperties(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, addFileFn func(path string, data string)) {
+func createOrUpdateCassandraRackConfig(rctx *reconciliationRequestContext, rack string) (*corev1.Volume, error) {
```
zegelin: Should the `rack` parameter be of type `Rack`?

pkg/controller/cassandradatacenter/configmap.go
alourie: yes, definitely.

pkg/controller/cassandradatacenter/configmap.go
alourie: switched to *Rack for all function calls.

smiklosovic:
merged here https://github.com/instaclustr/cassandra-operator/commit/c316b7416a0fa5a3442ea4e0fe13272ebef1ac70

Issue #228, opened 2019-08-09
Unused code in cassandradatacenter?


johananl:
It seems like the following code isn't used anywhere:

https://github.com/instaclustr/cassandra-operator/blob/84737bdf33c7b1b7d39732d626e47081550c84b1/pkg/controller/cassandradatacenter/ports.go#L12-L18

Can we get rid of this source file?

zegelin:
I think I was in the middle of trying to clean up and centralise port definitions between the container, the statefulset and the services. But it didn’t get finished.

PR #246, opened 2019-08-22
consolidated ports

smiklosovic:
should resolve https://github.com/instaclustr/cassandra-operator/issues/228

### DIFF
--- a/pkg/controller/cassandradatacenter/services.go
+++ b/pkg/controller/cassandradatacenter/services.go
@@ -17,15 +17,12 @@ func createOrUpdateNodesService(rctx *reconciliationRequestContext) (*corev1.Ser
 	opresult, err := controllerutil.CreateOrUpdate(context.TODO(), rctx.client, nodesService, func(_ runtime.Object) error {
 		nodesService.Spec = corev1.ServiceSpec{
 			ClusterIP: "None",
-			Ports: []corev1.ServicePort{
-				{Name: "cql", Port: 9042},
-				{Name: "jmx", Port: 7199},
-			},
-			Selector: DataCenterLabels(rctx.cdc),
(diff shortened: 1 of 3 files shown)


pkg/controller/cassandradatacenter/services.go
```diff
@@ -51,10 +48,8 @@ func createOrUpdateSeedNodesService(rctx *reconciliationRequestContext) (*corev1
 
 	opresult, err := controllerutil.CreateOrUpdate(context.TODO(), rctx.client, seedNodesService, func(_ runtime.Object) error {
 		seedNodesService.Spec = corev1.ServiceSpec{
-			ClusterIP: "None",
-			Ports: []corev1.ServicePort{
-				{Name: "internode", Port: 7000, Protocol: corev1.ProtocolTCP},
```
smiklosovic: TCP protocol is default

pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -127,12 +127,12 @@ func newCassandraContainer(cdc *cassandraoperatorv1alpha1.CassandraDataCenter, d
...
 		ImagePullPolicy: cdc.Spec.ImagePullPolicy,
 		Args:            []string{OperatorConfigVolumeMountPath},
-		Ports: []corev1.ContainerPort{
-			{Name: "internode", ContainerPort: 7000},
-			{Name: "internode-tls", ContainerPort: 7001},
-			{Name: "cql", ContainerPort: 9042},
-			{Name: "jmx", ContainerPort: 7199},
-		},
+		Ports: ports{
+			{"internode", 7000},
```
zegelin: Since these are used in multiple places, it might make sense to make them constants. Especially since now that we have a `port` type that can be converted to either a `Service...` or `ContainerPort` type, they can be kept in one spot.

PR #253, opened 2019-08-28
Improve reconcile loop

alourie:
The current reconcile loop is not going to work if the decommission or creating of the new nodes takes a long time. The events that the reconcile responds to will come in fast initially, and if the cluster is undergoing an operation, the operator will do nothing with those events.

This change fixes this issue by requesting a repeat of the handler call after 1 minute after detecting that the cluster is in a "non-ready state"; this way the handler will always be called until all the events are worked out on and the cluster is reconciled to the new requested state. This works for building, scaling up and scaling down scenarios.

### DIFF
--- a/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
+++ b/pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
@@ -2,7 +2,6 @@ package cassandradatacenter
 
 import (
 	"context"
-
 	"github.com/go-logr/logr"
 	cassandraoperatorv1alpha1 "github.com/instaclustr/cassandra-operator/pkg/apis/cassandraoperator/v1alpha1"
 	"k8s.io/api/apps/v1beta2"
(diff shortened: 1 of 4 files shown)


pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -122,8 +122,13 @@ func (r *ReconcileCassandraDataCenter) Reconcile(request reconcile.Request) (rec
 	}
 
 	statefulSet, err := createOrUpdateStatefulSet(rctx, configVolume)
-	if err != nil && err != ErrorCDCNotReady {
-		return reconcile.Result{}, err
+	if err != nil {
```
smiklosovic: What happens, in theory, when other event is triggered (for whatever reason) before one minute passes?

alourie: Nothing special, it will be put into the "queue" to be worked on. Basically, all events that trigger the handler are the same from our perspective, we don't know what initiated them, but they all become the same "request" with name and namespace.

So in this case, if there are 2 or more "competing" calls in the queue, the first one to get to a state where the cluster is "ready" will win and cause some action. Then the other one will be still in the queue, waiting until worked on, meaning that either there's some reconcile will be happening or nothing will happen (meaning nothing left to reconcile). Then the call will be cleared from the queue.

smiklosovic:
merged here https://github.com/instaclustr/cassandra-operator/commit/0037b8214659a49659086c613251aa26330f312d

PR #254, opened 2019-08-28
implementation of restores in go

smiklosovic:


docker/cassandra-sidecar/cassandra-backup-restore.sh
```diff
@@ -0,0 +1,36 @@
...
+        ;;
+    esac
+done
+
+if [[ "${args[0]}" == "restore" ]]; then
+    if [[ ! -f ${RESTORE_MARKING_FILE} ]]; then
+        echo "Executing restore command with arguments:" ${args[*]}
+        java -jar /opt/lib/cassandra-backup-restore/cassandra-backup-restore.jar ${args[*]} \
+        && touch ${RESTORE_MARKING_FILE} \
+        && echo "${args[*]}" > ${RESTORE_MARKING_FILE}
```
smiklosovic: this could be just echo without touching

java/backup-restore/src/main/java/com/instaclustr/cassandra/backup/cli/RestoreApplication.java
```diff
@@ -48,5 +48,9 @@ public void run() {
         final Operation operation = operationsService.submitOperationRequest(request);
 
         await().forever().until(() -> operation.state.isTerminalState());
+
+        if (operation.state == Operation.State.FAILED) {
+            throw new IllegalStateException("Restore operation was not successful.");
```
smiklosovic: this makes command return with exit code 1 so we can detect restore has failed in init container so whole init container will fail

alourie: :+1:

PR #258, opened 2019-08-29
Helm chart update

benbromhead:
Update Helm charts to support the new Go hotness!


smiklosovic:
merged here https://github.com/instaclustr/cassandra-operator/pull/260

PR #261, opened 2019-08-30
added logback.xml file to sidecar

smiklosovic:


smiklosovic:
https://github.com/instaclustr/cassandra-operator/issues/257

PR #262, opened 2019-08-30
Use Go module mirror

johananl:
This greatly reduces the time it requires for the Go tool to download modules.

More info: https://blog.golang.org/module-mirror-launch


smiklosovic:
https://github.com/instaclustr/cassandra-operator/commit/2bfbedcaaf36f4750d65486fb46dde52008a9a93

PR #265, opened 2019-08-31
Fixes and improvements

alourie:
Updates to reconcile loop:

* Refactor the checkClusterHealth to run before initiating the
reconcile.

This new step should considerably improve the workflow by checking the
cluster health before doing anything at all. If the cluster is not in an
ok state, the operator should create/update any additional resources and
should just wait until the cluster is "ok" again.

The additional benefit is that we don't need to check the health of the
cluster again during the specific StatefulSet reconcile call, which
significantly simplified the code.

* Refactor reconciliationContext to include all current stateful sets,
all pods and sidecar clients. Additionally it now includes the type of
the current scaling operation (up, down or none)

This allows building these resources once for any given reconciliation
cycle. They also become available immediately to lots of other functions
that already receive this context and make the code more readable.

* Small fixes and improvements.
* More comments explaining important areas.

Signed-off-by: Stefan Miklosovic <stefan.miklosovic@instaclustr.com>

### DIFF
--- a/pkg/common/cluster/utils.go
+++ b/pkg/common/cluster/utils.go
@@ -10,16 +10,14 @@ func BuildRacksDistribution(spec v1alpha1.CassandraDataCenterSpec) (racksDistrib
 	// If racks are not provided, we place everything in 1 Rack
 	if spec.Racks == nil {
 		spec.Racks = []v1alpha1.Rack{{Name: "rack1"}}
-	} else  {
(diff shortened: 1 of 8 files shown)


smiklosovic:
merged here https://github.com/instaclustr/cassandra-operator/commit/ae81eb94f690122ddb6077b53065433544666b88

PR #269, opened 2019-09-04
Make kernel tweaks optional

johananl:
## General

This PR makes kernel performance tweaks optional. These are controlled using the `optimizeKernelParams` field in the CRD.

### Default Behavior

By default, the C* pods created by the operator will run in the following mode:

- `sysctl` isn't called (the init container isn't included in the pod).
- `ulimit` isn't called.
- `disk_access_mode` is set to the default (`auto`).
- C* is unable to use the `mlock`/`mlockall` syscalls. A warning is written to the log.

### Kernel Tweaks Enabled

When the user sets `optimizeKernelParams` to `true`, the C* pods will run in the following mode:

- `sysctl -w vm.max_map_count=1048575` is executed in an init container. This changes the parameter **globally** on the host. Any related documentation should reflect that.
- `ulimit -l unlimited` is executed on C* startup. This change affects only the C* container.
- `disk_access_mode` is set to `mmap`. C* is able to lock memory.

## Implementation

The above is achieved with the C* container running without `privileged: true` and without running the process inside the container as root. This greatly reduces various security risks.

The above is implemented by conditionally using a Bash binary which has just the required capabilities for running `ulimit` and for performing memory locking. The reason I've implemented it this way:

- At the time of writing, setting Linux resource limits via k8s [isn't supported](https://github.com/kubernetes/kubernetes/issues/3595).
- A workaround is to call `ulimit` from within the container. To do so *without running as root*, the executed **binary** must have the `SYS_RESOURCE` capability. In our case that executable is Bash which is the interpreter used in the container's entrypoint.
- In order to be able to lock memory, a process needs the `IPC_LOCK` capability. To achieve that without running as root, we add that capability to the entrypoint's executable, similarly to the above.

### DIFF
--- a/docker/cassandra/Dockerfile
+++ b/docker/cassandra/Dockerfile
@@ -10,6 +10,13 @@ COPY install-cassandra /tmp/install-cassandra
 RUN /tmp/install-cassandra ${cassandra_version}
 
 COPY entry-point /usr/bin/entry-point
+COPY wrapper /usr/bin/wrapper
+
+# Allow entrypoint script to modify ulimit by creating a "patched" Bash executable
+# which is allowed to use the required capabilities
+RUN cp /bin/bash /bin/bash-mod \
+    && setcap cap_ipc_lock=+ep /bin/bash-mod \
+    && setcap cap_sys_resource=+ep /bin/bash-mod
 
 COPY cassandra /usr/sbin/cassandra
 COPY nodetool /usr/bin/nodetool
@@ -37,4 +44,4 @@ EXPOSE 7000 7001 7199 9042 9160
 # The user is created in ./install-cassandra
 USER 999
 
-ENTRYPOINT ["/usr/bin/entry-point"]
+ENTRYPOINT ["/usr/bin/wrapper"]
--- a/pkg/controller/cassandradatacenter/helpers.go
+++ b/pkg/controller/cassandradatacenter/helpers.go
@@ -81,3 +81,8 @@ func rackExist(name string, sets []v1.StatefulSet) bool {
 	}
 	return false
 }
+
+// boolPointer returns a pointer to bool b.
+func boolPointer(b bool) *bool {
+	return &b
+}
--- a/deploy/crds/cassandraoperator_v1alpha1_cassandradatacenter_crd.yaml
+++ b/deploy/crds/cassandraoperator_v1alpha1_cassandradatacenter_crd.yaml
@@ -52,7 +52,7 @@ spec:
             nodes:
               format: int32
               type: integer
-            privilegedSupported:
+            optimizeKernelParams:
               type: boolean
             prometheusSupport:
               type: boolean
--- a/docker/cassandra/entry-point
+++ b/docker/cassandra/entry-point
@@ -12,4 +12,9 @@ do
 done
 )
 
+# Set unlimited memory locking only when running with kernel tweaks enabled.
+if [ -n "${MEMORY_LOCK-}" ] && [ "${MEMORY_LOCK}" == "true" ]; then
+    ulimit -l unlimited
+fi
+
 exec /usr/sbin/cassandra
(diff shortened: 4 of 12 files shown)


docker/cassandra/Dockerfile
```diff
@@ -10,6 +10,13 @@ COPY install-cassandra /tmp/install-cassandra
 RUN /tmp/install-cassandra ${cassandra_version}
 
 COPY entry-point /usr/bin/entry-point
+COPY wrapper /usr/bin/wrapper
+
+# Allow entrypoint script to modify ulimit by creating a "patched" Bash executable
+# which is allowed to use the required capabilities
+RUN cp /bin/bash /bin/bash-mod \
```
benbromhead: This seems a little odd to me. Why are we doing ulimit modifications via a "patched" bash?

pkg/controller/cassandradatacenter/helpers.go
```diff
@@ -81,3 +81,8 @@ func rackExist(name string, sets []v1.StatefulSet) bool {
 	}
 	return false
 }
+
+// boolPointer returns a pointer to bool b.
+func boolPointer(b bool) *bool {
+	return &b
+}
```
alourie: :+1:

docker/cassandra/Dockerfile
smiklosovic: @benbromhead I think this solution originates from here (1) and distilled comments bellow it. Our "exe" is basically `bash` which is executing entry-point from wrapper script where Cassandra is started.

By using setcap below in this file, that "exe" (patched bash) can set ulimits (without root) which is what we want.

@johananl is that how it works for sure? Is my understanding correct?

(1) https://github.com/kubernetes/kubernetes/issues/3595#issuecomment-438507708

johananl: @benbromhead what @smiklosovic wrote is correct. I've tried to explain the rationale in the description but let me recap here:

- We want to let the C* process lock memory while running as a non-root user.
- We can't give C* the required Linux capabilities via k8s, which means we have to do it somehow from within the container.
- We want the above to be *conditional*, i.e. we should be able to also run without requiring the extra capabilities, for example when running on clusters with restrictive security policies.
- C* runs a Bash script on startup which tries to do `ulimit -l unlimited`. The **binary** which does so (which is `/bin/bash`) needs to have the `SYS_RESOURCE` capability. Additionally, C* needs to use `mlock` syscalls. For that, the **binary** used for spawning the C* process needs to have the `IPC_LOCK` capability. This binary would normally be the JVM, however since we `exec` the C* process from the entrypoint script (rather than creating a child process), we can set the `IPC_LOCK` capability on the Bash binary as well.
- The reason we need to have both a Bash binary with and without the capabilities is that when we run in "low privileged" mode (i.e. without these tweaks), the C* container won't start if `/bin/bash` has the extra capabilities because the PSPs won't allow it.

On 2nd thought we *may* be able to achieve what we want with a **patched JVM**, however this would only solve the problem for `IPC_LOCK` since `SYS_RESOURCE` is still needed by `/bin/bash`.

smiklosovic: thanks @johananl for the confirmation, we will go with this solution for now. I ll merge it in a while after some testing on my side.

johananl: Great. Please do test it thoroughly as I may have missed something, especially things like unexpected Cassandra behavior in either of the modes ("normal" / "privileged").

johananl: >On 2nd thought we may be able to achieve what we want with a patched JVM, however this would only solve the problem for IPC_LOCK since SYS_RESOURCE is still needed by /bin/bash.

For the record, I've looked into this and it doesn't seem to be possible because it's not the Java process which runs `ulimit -l unlimited`. Even if we run the `ulimit` command [here](https://github.com/instaclustr/cassandra-operator/tree/superdupertopsecretrewrite/docker/cassandra/default-config/cassandra-env.sh.d) it will still be executed by `/bin/bash`.

PR #274, opened 2019-09-06
Fix CI

benbromhead:
This PR gets CI working again. However only in a somewhat basic and brittle manner. Currently CI works in the following manner:

- Make is run for all PRs to show that the project builds
- On merge to `superdupertopsecretrewrite` the project will be built and docker images published with the *-go tag affixed 
- Integration tests in GCP will be run
- There are a lot of hardcoded names with respect to image names etc. These will need to be fixed on the flip up to master - or if someone just wants to make this whole thing slightly more resilient (not sure the ROI on doing this right now tho until after this is the main workin branch).

Currently integration tests are done the old java way and don't make any sense. Next steps is to get test/e2e working against the GCP cluster. 

This is really just a building block that runs unit tests, some integration tests and a way for getting build images out for broader testing!


smiklosovic:
it is in! https://github.com/instaclustr/cassandra-operator/commit/24d2ae8e094194d8b340007ff74ee83988a61fb0

PR #277, opened 2019-09-08
Add support for tolerations and affinity

sukrit007:
Adds support for passing tolerations and affinity to the pod spec inside stateful set .
It also includes fixes for java image

**Purpose**
- With affinity and anti-affinity, we can target Cassandra nodes to be distributed across Kubernetes nodes. No 2 pods 
- With tolerations, we can target Cassandra nodes to be run on nodes dedicated to Cassandra.

### DIFF
--- a/docker/base-openjre/Dockerfile
+++ b/docker/base-openjre/Dockerfile
@@ -1,6 +1,6 @@
 FROM debian:stretch
 
-ARG openjre_version=8u212-b01-1~deb9u1
+ARG openjre_version=8u222-b10-1~deb9u1
 
 COPY dagi /usr/local/bin/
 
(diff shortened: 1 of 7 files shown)


smiklosovic:
@sukrit007 thanks! It is in master.

I would like to let you know that we are working very hard towards Go version of operator build on top op operator-sdk framewor and Java version of operator will not be actively maintained after this switch is done (matter of weeks).

sukrit007:
@smiklosovic   Is this something we can add it to go version? If needed I can send a PR for the same.

smiklosovic:
@sukrit007 yes you are welcome to do so but could you please wait until that transition is done? we are having Go operator in a branch and we need to promote it to master first ... should take a week or two ...

PR #279, opened 2019-09-11
fixed Helm charts

smiklosovic:
I deleted old java-based cassandra-operator builds scripts too.

~I deleted zipped charts after they are installed and put them into gitignore, same with index. We should not commit any binary blobs into repository.~


helm/cassandra/values.yaml
```diff
@@ -16,87 +18,82 @@ image:
...
+#  - name: configMapReference
 
+#restoreFromBackup: backup-name
 
-#restoreFromBackup: backup-hostname
+prometheusEnabled: false
 
-prometheusEnabled: false
\ No newline at end of file
+#serviceAccountName: cassandra
```
smiklosovic: @johananl should not this be enabled? wdyt? I think when Cassandra operator is deployed, it will deploy all service accounts and roles along with that so this should be fine to have not commented out, right?

johananl: I think it is probably safer to explicitly use that SA, yes. On most k8s clusters things will probably work also with the default SA (now that we don't run as root), however on *some* clusters with tighter PSPs it will break. So yes.

PR #292, opened 2019-09-19
Parametrize build scripts

mmonemali:
Unfortunately we work behind proxy and we don't have the luxury of connecting directly to the internet or use many of the public repositories.
To get around that this PR just made all build Makefiles/Dockerfiles configurable via environment variables, all the current existing defaults are maintained.

> export C_APACHE_MIRROR_URL=
> export CASSANDRA_OPERATOR_VERSION=
> export OPENJRE_VERSION=
> export OPENJRE_BASE_IMAGE=
> export GOLANG_BASE_IMAGE=
> export CASSANDRA_VERSION=
> export INSTALL_CASSANDRA_EXPORTER=


smiklosovic:
interesting PR, thanks for that @mmonemali !

@alourie @johananl what do you think?

docker/cassandra/install-cassandra
```diff
@@ -39,10 +42,13 @@ mkdir /etc/cassandra/cassandra-env.sh.d
...
 
-# install cassandra-exporter (Prometheus monitoring support)
-(cd "/usr/share/cassandra/agents" &&
-    curl -SLO "https://github.com/instaclustr/cassandra-exporter/releases/download/v0.9.6/cassandra-exporter-agent-0.9.6.jar" &&
-    ln -s cassandra-exporter-agent-0.9.6.jar cassandra-exporter-agent.jar)
+if [ "$INSTALL_CASSANDRA_EXPORTER" == "true" ]
+then
+    # install cassandra-exporter (Prometheus monitoring support)
+    (cd "/usr/share/cassandra/agents" &&
+        curl -SLO "https://github.com/instaclustr/cassandra-exporter/releases/download/v0.9.6/cassandra-exporter-agent-0.9.6.jar" &&
```
smiklosovic: @mmonemali I would parametrize also this version 0.9.6 if it is possible.

alourie: 0.9.8 is the latest ;-)

alourie:
Thanks for the contribution, @mmonemali! It does look good to me. Let's parametererise the exporter version as well, and we're golden.

mmonemali:
Thanks for the review guys, I did make the exporter version also configurable. But I think it is better to increase the version in a separate PR?

smiklosovic:
> Thanks for the review guys, I did make the exporter version also configurable. But I think it is better to increase the version in a separate PR?

@mmonemali thanks for that! Why we could not increase the version in this very same PR?

mmonemali:
> > Thanks for the review guys, I did make the exporter version also configurable. But I think it is better to increase the version in a separate PR?
> 
> @mmonemali thanks for that! Why couldn't we increase the version in this very same PR?
> 
> edit: ah, I think you are right with separate PR. We'll get to it after this sneaks in ...

Yeah just for whatever reason if needed to revert the version change or my changes they will be separate:)

PR #294, opened 2019-09-20
Support exporting Prometheus metrics

### LINKED ISSUE #180: Port Prometheus support
The java based operator supports configuring Cassandra to expose metrics in a prometheus friendly manner (see instaclustr/cassandra-exporter). The operator also supports automatic discovery of new clusters and nodes within prometheus.

We need to ensure the operator-sdk versions supports this workflow

alourie:
Fix #180 

This will create ServiceMonitor object for Cassandra if prometheus-operator is installed in this k8 environment. It is expected that the Prometheus resource is going to be created outside of our operator, and lables will be provided to us.


doc/custom-configuration.md
```diff
@@ -91,3 +91,38 @@ spec:
 
 Cassandra will load the `cassandra.yaml.d/100-concurrent.yaml` file as well as the default settings managed by the operator!
 
+### Prometheus support
+
+The Cassandra image that is included with this project has the [cassandra-exporter](https://github.com/instaclustr/cassandra-exporter) built in, so it is ready to export the metrics to Prometheus instance. In order to use this capability automatically, one must have a running Prometheus operator in the same kubernetes environment and have Prometheus object created bound to a set of ServiceMonitors via a selector (a set of labels); for example:
```
smiklosovic: Kubernetes environment

smiklosovic: "created bound" ?

pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
```diff
@@ -162,6 +162,13 @@ func (r *ReconcileCassandraDataCenter) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
+	if rctx.cdc.Spec.PrometheusSupport {
+		err := createOrUpdatePrometheusServiceMonitor(rctx)
```
smiklosovic: could not you do `if err := _that_method; err != nil { return reconcile.Result{}, err` ?

pkg/controller/cassandradatacenter/metadata.go
```diff
@@ -20,6 +20,16 @@ func DataCenterLabels(cdc *cassandraoperatorv1alpha1.CassandraDataCenter) map[st
 	}
 }
 
+func PrometheusLabels(cdc *cassandraoperatorv1alpha1.CassandraDataCenter) map[string]string {
+	// Fetch cdc labels
+	labels := DataCenterLabels(cdc)
+	// Add prometheus labels if defined
+	for label, val := range cdc.Spec.PrometheusServiceMonitorLabels {
+		labels[label] = val
```
smiklosovic: do we really need to propagate all labels from `DataCenterLabels(cdc)` to Prometheus operator? Isn't just an empty map to start with fine?

pkg/controller/cassandradatacenter/cassandradatacenter_controller.go
alourie: I did it the same way we run other `createOrUpdate...`, but sure, why not.

pkg/controller/cassandradatacenter/metadata.go
alourie: Yes, we do need the CDC labels here. The Prometheus service that we create (`cassandra-<cdc name>-prometheus`, which is where it is used) selects pods via labels, and that's where we need those labels - in the service selector. Then the `ServiceMonitor` has a selector to pick the service for monitoring, and it will use the same set of labels to pick a service for the specific CDC; then the Prometheus resource (which we don't care for nor create) will use its own selector to find ServiceMonitors.

We don't propagate our CDC labels to Prometheus operator. These labels are for our internal use and for the `ServiceMonitor` that we create.

doc/custom-configuration.md
alourie: Fixed both.

pkg/controller/cassandradatacenter/services.go
```diff
@@ -79,3 +113,35 @@ func createOrUpdateSeedNodesService(rctx *reconciliationRequestContext) (*corev1
...
 	return seedNodesService, err
 }
+
+func createOrUpdatePrometheusServiceMonitor(rctx *reconciliationRequestContext) error {
+	cfg, err := config.GetConfig()
+	if err != nil {
+		return err
+	}
+
+	prometheusService, err := createOrUpdatePrometheusService(rctx)
```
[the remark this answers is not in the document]
alourie: yea, that's the main issue. I can make it work with the current `-nodes` service, but the presentation would be ugly on the UI, as you'd have other ports "failing" and such.

doc/custom-configuration.md
```diff
@@ -91,3 +91,38 @@ spec:
 
 Cassandra will load the `cassandra.yaml.d/100-concurrent.yaml` file as well as the default settings managed by the operator!
 
+### Prometheus support
+
+The Cassandra image that is included with this project has the [cassandra-exporter](https://github.com/instaclustr/cassandra-exporter) built in, so it is ready to export the metrics to Prometheus instance. In order to use this capability automatically, one must have a running Prometheus operator in the same Kubernetes environment and have Prometheus object following a set of ServiceMonitors via a selector (a set of labels); for example:
```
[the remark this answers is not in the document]
alourie: Yea, I can agree that we don't need to explain how to deploy Prometheus. The reason I did it here, is because for me, as I never actually ran Prometheus before, it was somewhat complicated to understand the flow of things. So I'll remove it for now, but if people ask for clarification, I reckon we can either put it back or at least reference the official Prometheus docs.

Also, the fact that labels put on _some_ service are _also_ put on the `ServiceMonitor` is purely circumstantial, and users definitely shouldn't have to be aware of this detail of the implementation. The CRD parameter is called `prometheusServiceMonitorLabels` for a reason - it's the set of labels used in their own selector in the Prometheus object (which they still can configure any way they want), and that's how they should think about it. It's the set of labels that is going to be used to label the ServiceMonitor we create so that Prometheus picks it up.

pkg/controller/cassandradatacenter/services.go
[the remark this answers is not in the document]
alourie: and I will open an issue upstream to make it more flexible. Thanks!

doc/custom-configuration.md
```diff
@@ -91,3 +91,38 @@ spec:
...
+```
+
+`prometheus` service account used above is taken literally from the Prometheus' [user guide](https://github.com/coreos/prometheus-operator/blob/master/Documentation/user-guides/getting-started.md#enable-rbac-rules-for-prometheus-pods). Use the example above to create Prometheus object that uses a specified set of ServiceMonitor labels.
+
+  When Prometheus object is created, and the labels are at hand, add the following configuration to the cluster CRD:
+```yaml
+...
+spec:
+ ...
+  prometheusServiceMonitorLabels:
```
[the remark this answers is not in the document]
alourie: I disagree with this. As mentioned in my [comment](https://github.com/instaclustr/cassandra-operator/pull/294#discussion_r328405381), these labels _are_ specific to the Prometheus support.  These labels come from the Prometheus' selector for the ServiceMonitors, they are needed so that Prometheus picks up the ServiceMonitor _we_ create. These specific labels have nothing to do with our internal services, which we create for the Cassandra discovery. We only use them all the way Prometheus->ServiceMonitor->Service because it's convenient and easy to debug if there are issues. We could have a constant label such as `prometheus: <cdc name>` to uniquely pick the service from the ServiceMonitor, without asking users for labels, if we didn't have to integrate with Prometheus itself.

That being said, I can definitely see a benefit in the ability to add arbitrary labels to any node, service, StatefulSet or pod created by the operator. In theory, it could overlap/encompass this feature by providing a more generic labelling mechanism. I've created a separate ticket #299 for that, as it would be way more expansive than just for this feature.

alourie:
I will open a separate ticket for the upstream. If/when it's implemented, we'd be able to simplify this by using just 1 service (`-nodes`) instead of creating an additional `-prometheus` one.

doc/custom-configuration.md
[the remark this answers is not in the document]
johananl: What I meant was that we don't need a special field for "Prometheus-related labels", we can achieve the same end result with a generic service labels field since in any case this is up to the user to figure out what kind of selector (if any) they have in their Prometheus config and apply matching labels to the service. But anyway, feel free to disregard this suggestion :-)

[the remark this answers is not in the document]
alourie: :+1:

At the moment, until we implement some kind of "generic" labelling, I prefer keeping the name specific to Prometheus, as that would be more clear to the users...IMHO, of course.

PR #298, opened 2019-09-24
Added more checks in CI

alourie:
Additional "repo" checks in CI will validate that OpenAPI and k8s
objects are in sync with the code changes; as well as that `make bundle`
was rerun and is also in sync with the current code state.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,33 +3,23 @@
 #
 version: 2
 jobs:
-  style-check:
+  repo-check:
     docker:
       - image: circleci/golang:1.13.0
 
-    working_directory: ~/cassandra-operator
+    working_directory: /go/src/github.com/instaclustr/cassandra-operator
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -3,33 +3,23 @@
...
       - checkout
       - restore_cache: # restores saved cache if no changes are detected since last run
           keys:
             - go-mod-v4-{{ checksum "go.sum" }}
       - run:
-          name: Update go stuff
-          command: go get golang.org/x/tools/cmd/goimports && go mod download
-
+          name: Validate repository
+          command: GO111MODULE=on .circleci/validate_repo.sh
```
smiklosovic: isnt this GO111MODULE=on already default in 1.13?

alourie: yes, but for some reason, it doesn't work without it being set.

PR #305, opened 2019-10-02
Documentation update

### LINKED ISSUE #304: StatefulSet PVC are not zone aware
I tried out the zone-aware configuration on AWS and it failed because volumes were created in the wrong zone.

Kubernetes version: `v1.14.6`
cassandra-operator: `v2.0.0`

The non-zone aware configuration worked fine. It put 2 out of 3 Pods in the same zone but it worked.

The zone-aware setup looks more robust so I tried the following config:
```yaml
  nodes: 3
  racks:
  - name: eu-central-1a
    labels:
      failure-domain.beta.kubernetes.io/zone: eu-central-1a
  - name: eu-central-1b
    labels:
      failure-domain.beta.kubernetes.io/zone: eu-central-1b
  - name: eu-central-1c
    labels:
      failure-domain.beta.kubernetes.io/zone: eu-central-1c
```

The node in zone `1a` spawned fine but nodes in `1b` and `1c` couldn't be scheduled due to zone volume conflicts. The PV for the `1b` PVC landed in zone `1a`.

I also tried using `nodeAffinity` instead of `nodeSelector` on the cassandra `StatefulSet` but it didn't help.

After talking to a collegue this seems to be a shortcoming of the `StatefulSet` controller itself and not of `cassandra-operator`.

One way to fix it here would be to pre-create three individual `StorageClass`es for each zone. We could then add a key to the configuration above defining which storage class to use for each rack configuration, e.g.
```yaml
  nodes: 3
  racks:
  - name: eu-central-1a
    labels:
      failure-domain.beta.kubernetes.io/zone: eu-central-1a
      storage-class: default-1a
    ...
```

What are your thoughts about this?

Your examples use zone configurations for Google Cloud. Have you seen theses problems in your testing with the zone aware configuration?

alourie:
Closes #304 

Update the documentation for dynamic PV provisioning: rename the doc
from GKE-specific to pv-generic.


smiklosovic:
@alourie what about Helm charts? Could you maybe update them as well somehow to have this covered? If there are rack-related Helm bits, this should be there somehow too. Otherwise, all good!

alourie:
@smiklosovic the thing is that this parameter is optional (and rack as well), so I don't know if we need to provide anything there for the helm charts. Should we add them as comments maybe? This way they are not enabled but appear in the template, similar to what Linux configs do in many cases.

smiklosovic:
yes, commenting it makes sense

linki:
Thanks for pointing me to this. lgtm

PR #306, opened 2019-10-08
Cassandra has many names

### LINKED ISSUE #57: Tag `cassandra` image with Cassandra version
Curently, `build` & `build-all` tag images with the repo short hash & `latest`. This is useful for development.

Ideally the Cassandra image should be tagged with the version of C* it represents, e.g. `3.2.11`.
This makes it obvious what version is contained within, and will be necissary once we need to support multiple C* versions.

alourie:
Closes #57 

### DIFF
--- a/buildenv/build-all
+++ b/buildenv/build-all
@@ -16,5 +16,7 @@ if [[ "${CIRCLE_BRANCH}" == master ]]; then
     docker tag cassandra-sidecar:latest "${REGISTRY}/cassandra-sidecar:${TAG}"
     docker tag cassandra-sidecar:latest "${REGISTRY}/cassandra-sidecar:latest"
     docker tag cassandra:latest "${REGISTRY}/cassandra:latest"
+    docker tag cassandra:latest "${REGISTRY}/cassandra:${TAG}"
     docker tag cassandra:3.11.4 "${REGISTRY}/cassandra:3.11.4"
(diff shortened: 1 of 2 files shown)


smiklosovic:
@alourie this means that everytime we change something in Cassandra image, even there are not any other changes elsewhere, we will be forced to do a release, @benbromhead @zegelin are we ok with that?

alourie:
@smiklosovic Yea, I'd agree, but remember that we do a lot of customisations in there _for_ the operator specifically. We're not just wrapping the Cassandra itself, we're actually doing some operator-specific thing....unless we don't.

Either way, how would you suggest tagging this image then? We currently have no other mechanism to release our container images, so if there is any other idea, I'm all for it :-)

mmonemali:
I'm just curious what things in the operator depends on the Cassandra version that is running, and if these things can be eliminated/refactored somehow.

smiklosovic:
@mmonemali virtually none. The whole communication to Cassandra is done only via its sidecar container via JMX. Some method _might_ change in JMX across versions but the differences should be negligible if any. If switch is done between 3.x versions, it should be just fine. Not sure if method signatures for JMX in 4.0 differ ... Apart from this, it should be fine. Maybe some configuration options for JVM? I can imagine that can be different little bit e.g. the configuration of garbage collectors etc ... Some options can be replaced by fragment mechanism already.

PR #311, opened 2019-11-07
Fix rack naming

mmonemali:
Issue: If you create datacenter with minimal configs without specifying racks the operator will put everything in one rack named 'rack1', if you later create another datacenter also without specifying racks also everything will be put in rack1 and while checking if we need to reconcile any racks the 2nd datacenter won't be created because rack1 is up and running from the previous datacenter.

Fix: if racks are not specified, put everything in one rack named cdc.Namespace + cdc.Name + rack so we have unique rack names for different datacenter.


smiklosovic:
Thank you @mmonemali , I ll take a look in a while. Do not hesitate to mention somebody in your PRs for having a look faster.

mmonemali:
@smiklosovic @alourie  Thanks! I think I'll actually close this PR for now, probably we need to shorten the rack name even more because it is used as a label as well and because of the 63 length limit on labels pods would fail to schedule if the cluster name + namespace name is long enough > 63

The idea is to give it a non random name that could be regenerated on each reconciliation

smiklosovic:
@mmonemali just hash that with md5 to have it of constant length. It will be different on each namespace + name combination and shorter than 63 characters.

Also, can you please make your pull request in a such fashion that you create a new branch (naming it by issue it solves) and apply your commits on top of current master? It gets complicated quite quickly when I am reviewing your master branch where a lot of other commits are somehow too (probably as a result of merging our master into that and so on ..)

PR #333, opened 2020-01-25
rewritten seeds resolution, nodes are streaming data on bootstrap

smiklosovic:


pkg/controller/cassandradatacenter/statefulset.go
```diff
@@ -132,7 +132,7 @@ func newStatefulSetSpec(
 	podRackLabels := RackLabels(cdc, rack)
 	podLabels := PodTemplateSpecLabels(cdc)
 	statefulSetSpec := &v1beta2.StatefulSetSpec{
-		ServiceName: "cassandra", // TODO: correct service name? this service should already exist (apparently)
+		ServiceName: "cassandra-" + cdc.Name + "-nodes",
```
smiklosovic: without this, we could not even ping other pods from the pod so hostname resolution was not working. `ServiceName` field has to be same as the nodes service. This is a common mistake people do and they wonder why they cant ping stuff.

java/k8s-addons/src/main/java/com/instaclustr/cassandra/k8s/SeedProvider.java
```diff
@@ -23,12 +26,40 @@ public SeedProvider(final Map<String, String> args) {
...
+            //
+            // After that, seed nodes will be only the first nodes in a rack,
+            // host names start with the name of a respective pod and the first node
+            // in a rack always ends with "-0" suffix so we select only these.
+            final List<InetAddress> seeds = nodes.stream()
+                                                 .map(address -> {
+                                                     try {
+                                                         return InetAddress.getAllByName(address.getHostAddress());
+                                                     } catch (Exception ex) {
+                                                         return null;
```
zegelin: Is silently ignoring exceptions the best thing to do?

java/k8s-addons/src/main/java/com/instaclustr/cassandra/k8s/SeedProvider.java
```diff
@@ -23,12 +26,40 @@ public SeedProvider(final Map<String, String> args) {
...
+            //
+            // We iterate over the list and resolve hostname of each such IP.
+            // This step will return host names like:
+            //
+            // cassandra-test-dc-cassandra-west1-a-0.cassandra-test-dc-cassandra-nodes.default.svc.cluster.local
+            //
+            // After that, seed nodes will be only the first nodes in a rack,
+            // host names start with the name of a respective pod and the first node
+            // in a rack always ends with "-0" suffix so we select only these.
+            final List<InetAddress> seeds = nodes.stream()
```
zegelin: This probably would just be easier to follow if it was just a regular for-loop rather than a stream pipeline.

pkg/controller/cassandradatacenter/services.go
```diff
@@ -62,56 +62,17 @@ func createOrUpdateNodesService(rctx *reconciliationRequestContext) (*corev1.Ser
...
-		seedNodesService.Spec = corev1.ServiceSpec{
 			ClusterIP:                "None",
-			Ports:                    internodePort.asServicePorts(),
+			Ports:                    ports{cqlPort, jmxPort, internodePort}.asServicePorts(),
 			Selector:                 DataCenterLabels(rctx.cdc),
 			PublishNotReadyAddresses: true,
 		}
 
-		seedNodesService.Annotations = map[string]string{
+		nodesService.Annotations = map[string]string{
```
zegelin: The (subtle) difference between the SeedNodesService and NodesService was that the former will resolve unready nodes, where as the latter will only resolve nodes that are up (responding to CQL requests).

By combining the two and always having it set `tolerate-unready-endpoints` to true means that apps can discover unready nodes.

java/k8s-addons/src/main/java/com/instaclustr/cassandra/k8s/SeedProvider.java
zegelin: This would make dealing with exceptions easier too.

pkg/controller/cassandradatacenter/services.go
smiklosovic: @zegelin why would you want to wait for unready endpoints anyway?

PR #337, opened 2020-02-03
Beta

benbromhead:
Let's look to move this to Beta status. 

I'm happy that the CRDs we have are more or less pretty stable and we have plenty of folk building and running good dev environments against this.

Unless we need to do anything more from a TLS perspective?

cc @zegelin @smiklosovic


smiklosovic:
@benbromhead wait a minute I am in the middle of releasing stuff

JensRantil:
Any progress here?

smiklosovic:
@JensRantil yes! there is ... we are releasing Icarus and Esop these days and right after that beta will happen :) Thank you for your interest.

PR #350, opened 2020-03-05
Use correct guard for missing annotations to avoid panic

linki:
The guard to avoid a `nil` panic checks `operatorLabels` instead of `operatorAnnotations`.


smiklosovic:
@linki thanks!

Issue #351, opened 2020-03-10
consolidete Sidecar and backup-restore into one JAR


smiklosovic:
Because Sidecar is using backup-restore's API for taking a backup and Sidecar image bundles backup-restore jar as it is used upon restores, we are effectively depending on backup-restore jar twice which makes whole image unnecessarilly bloated.

Make sidecar image built in such a way that backup-restore is embedded into Sidecar application so one Sidecar JAR can be used both as a server application as well as a CLI application in restoration container.

PR #352, opened 2020-03-10
merged backup-restore and sidecar jars together

smiklosovic:


smiklosovic:
solves https://github.com/instaclustr/cassandra-operator/issues/351

PR #375, opened 2020-05-04
added watch_namespace to values.yaml

senax:
This change will allow setting the 'WATCH_NAMESPACE' behaviour from values.yaml so that it can be used with overrides, without having to edit the helm chart.

Thanks,
Frank

### DIFF
--- a/helm/cassandra-operator/Chart.yaml
+++ b/helm/cassandra-operator/Chart.yaml
@@ -14,4 +14,4 @@ maintainers:
   - name: Ben Bromhead
   - name: Adam Zegelin
 name: cassandra-operator
-version: 1.1.0
+version: 1.1.1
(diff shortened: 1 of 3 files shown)


smiklosovic:
Hi @senax,

since you have changed Cassandra operator Helm chart, would you mind to update its version in `helm/cassandra-operator/Chart.yaml` to `1.1.1` ?

These charts are then released into GCP Docker registry (1) so you can pull them by Helm so it would be awesome if we stick with versioning.

Thanks a lot!

https://console.cloud.google.com/gcr/images/cassandra-operator/GLOBAL/cassandra-operator-chart?gcrImageListsize=30

senax:
Sorry, forgot the chart version, should be bumped now.

PR #392, opened 2020-09-10
fixed tokens on restore and fixed rpc_address and broadcast_rpc_address

smiklosovic:
…ss for port forwarding


smiklosovic:
fixes https://github.com/instaclustr/cassandra-operator/issues/390

Issue #407, opened 2021-03-20
[BUG] Reading stale pod information can lead to undesired PVC deletion


srteam2020:
**Describe the bug**
We find that reading stale pod information from the apiserver will make the controller accidentally delete the PVC used by the current pod. More concretely, if we scale down and scale up a CassandraDataCenter, the controller (after a restart) may read a stale view of the pods and believe one of the cassandra pod is going to be deleted (due to the previous scale down). The controller will try to delete the PVC used by the "to-be-deleted pod", while the PVC is actually used by the current Cassandra pod. Currently, the controller lists the PVC using the `datacenter`, `datacenterUID`, and `cluster`. All three fields remain unchanged after scaling down/up, so the controller is not able to differentiate between the PVC used by the "to-be-deleted pod" and the current pod. One potential solution is to use the pod UID to list the PVC, so that when seeing stale pod information the controller will never try to delete the PVC held by the current pod.

This issue is a little bit similar to https://github.com/instaclustr/cassandra-operator/issues/402, but the solution PR https://github.com/instaclustr/cassandra-operator/pull/403 (list PVC using CassandraDataCenter UID) can only help to differentiate between CassandraDataCenters that share the same name when reading stale information. As mentioned above, using pod UID to differentiate between pods that share the same name can help here.

**To Reproduce**
The issue happened in a HA k8s cluster:
1. Create a CassandraDataCenter `cdc` with `nodes=2` and `deletePVCs=true`. There will be two cassandra pods `ca1` and `ca2` now. The controller is talking to apiserver1 now.
2. Scale `cdc` down (by setting `nodes=1`). `ca2` and its PVC will be deleted. Meanwhile, apiserver2 gets partitioned so its watch cache stops at the moment that `ca2` is tagged with a deletion timestamp.
3. Scale `cdc` up (by setting `nodes=2`). Now a new `ca2` and its PVC is back. Note that the new `ca2` shares the same name as the previously deleted one, so as its PVC.
4. After experiencing a crash, the restarted controller talks to the stale apiserver2. From apiserver2's watch cache, the controller finds that `ca2` is tagged with a deletion timestamp. The controller cannot differentiate between the previous PVC (which is already deleted) used by the old `ca2` and the existing one. As a result, the controller will delete `ca2`'s PVC, which is unexpected behavior.

**Expected behavior**
[...]

PR #408, opened 2021-03-24
Label each pvc with the pod UID

srteam2020:
Try to solve #407 
Label each pvc with the corresponding pod uid, and check the label before deleting the pvc


smiklosovic:
thanks!

PR #419, opened 2025-01-25
Update crds to use `apiextensions.k8s.io/v1`

adomenech73:
When applying the CRDs I am getting

```txt
error: resource mapping not found for name: "cassandraclusters.cassandraoperator.instaclustr.com" namespace: "" from "cassandra/deploy/crds/cassandraoperator.instaclustr.com_cassandraclusters_crd.yaml": no matches for kind "CustomResourceDefinition" in version "apiextensions.k8s.io/v1beta1"
ensure CRDs are installed first
```

The error is due to using an outdated API version for the CustomResourceDefinition (CRD). The apiextensions.k8s.io/v1beta1 version is no longer supported in newer Kubernetes versions.

To resolve this issue, I update the CRD to use the `apiextensions.k8s.io/v1` API version.


smiklosovic:
This repository is not supported by us anymore.

